PR #56, opened 2021-11-19
Search API enpoint

damisul:
Implemented /api/v1/search enpoint

In order to do this I made some additional changes:
- added some fields to ManifestationIndex
- changed format or download_url in response from direct link to Rails Active Storage, to /download/{id} url: it allows us to avoid fetching db data for Manifestation
- added genres validation to Work and Expression models
- made some minor changes to Models to avoid code duplicaiton

### DIFF
--- a/test/api/v1/api_test.rb
+++ b/test/api/v1/api_test.rb
@@ -6,9 +6,12 @@ class V1::APITest < ActiveSupport::TestCase
   def setup
     @key = create(:api_key)
     @disabled_key = create(:api_key, status: :disabled)
-    @manifestation = create(:manifestation, title: '1st', markdown: 'Sample Text 1')
-    @manifestation_2 = create(:manifestation, title: '2nd', markdown: 'Sample Text 2')
-    @unpublished_manifestation = create(:manifestation, status: :unpublished)
+
+    Chewy.strategy(:atomic) do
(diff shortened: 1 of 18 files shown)


test/api/v1/api_test.rb
```diff
@@ -180,6 +179,96 @@ def app
...
+    assert last_response.successful?
+    json = JSON.parse(last_response.body)
+    assert_equal 2, json['total_count']
+    data = json['data']
+    assert_equal 2, data.size
+    assert_manifestation(data[0], @manifestation_2, 'epub', false)
+    assert_manifestation(data[1], @manifestation, 'epub', false)
+  end
+
+  test 'GET /v1/api/search applies filters' do
```
abartov, line 213: where is the actual test here?

abartov (commented):
Looks good. Please see my comment about that one test.

test/api/v1/api_test.rb
damisul: Oops, fixed.

I started to writing this test and later switched to test for SearchManifestation service.

I've added couple of tests for API, one of the tests uses Mocha mock object simply to ensure that all supported parameters are correctly parsed and passed to SearchManifestation service correctly.

Also I've rebased branch on latest master.

abartov (approved):

PR #66, opened 2021-11-23
Moved linktype and link_status enums declaration from Manifestation t…

damisul:
Moved linktype and link_status enums declaration from Manifestation to ExternalLink model.
Refactored existing code to reflect those changes. Replaced manually added scopes with automatically-generated.

@abartov , I've tested most of changes, except manifestation#update and html_file#create_WEM_new, as I'm not sure how to find those pages in UI. Could you please ensure it is working?


abartov:
Now fixed.

PR #69, opened 2021-11-25
Enrichment view and snippet param

damisul:
Added Enrichment section to API response. Added separate param to manage snippet visibility.

For enrichment data I've decided not to add them into elastic search index, but fetch them from db during request.
The main reason for it is that if we want to include data in ES index it will require us to update index every time one of dependent records will be updated.

E.g. if we'll add or delete aboutness record will need to update manifestation index for referenced manifestation.
Same for recommendations, tags and external links

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,11 +95,11 @@ group :test do
   gem 'turn', '0.8.2', require: false
   gem 'simplecov', require: false
   gem 'factory_bot_rails', '~> 6.2.0', require: false
+  gem 'faker', '~> 2.19.0'
   gem 'mocha', '~> 1.13.0'
 end
 
 group :development do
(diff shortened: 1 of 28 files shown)


abartov (commented):
Looks good overall.  One change is needed, though:

The "works_about" should return *manifestation* IDs.  It's confusing, I realize, but the "texts" we return through the API are always and only Manifestations.  Works is an internal metadata entity of ours, which we don't expose, for now.  So once you retrieve the work_ids that are about this manifestation's work, you need to return the *manifestation IDs* associated with those works, not the work IDs themselves.

damisul:
Fixed. Also added test for Work.works_about

abartov (approved):

PR #89, opened 2021-12-15
API adjustements

damisul:
Removed fields publication_year and title_and_authors from Texts API response object.
Splitted field place_and_publisher into two fields: publication_place and publisher.


damisul:
resolved conflicts

PR #120, opened 2022-02-04
BaseUser bookmarks and preferences

damisul:
Modified Bookmarks and Preferences to be linked to BaseUser model instead of User model.
Allowed not-registered users to create bookmarks and edit preferences.
Modified ApplicationController.base_user method, to only create new BaseUser record if it is required explicitely

P.S. I see that only two preferences are used from code: fontsize and jump_to_bookmarks, but I couln't find places in UI where fontsize is set. Is it actually disabled?


abartov:
(Yes, only the jump_to_bookmarks preference is currently meaningful, but we will have several more preferences in the not-far future.)

PR #131, opened 2022-03-26
Dropped Expressions.genre column

damisul:
Dropped expressions.genre column, as we have it in works table as well.
Added specs for all affected places.

see https://github.com/abartov/bybeconv/issues/126


damisul:
Ok to test

abartov (approved):
Great work, thank you!

PR #144, opened 2022-05-13
Modified works <-> expressions relation to be one-to-many instead of many-to-many

damisul:
Replaced expressions_works table with column expressions.work_id

Updated all places where it is used

Added specs for most of affected places.

Also I've updated rake tasks, but I'm not sure how to test them.

See https://github.com/abartov/bybeconv/issues/135

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -2,7 +2,7 @@ include BybeUtils
 class Work < ApplicationRecord
   GENRES = %w(poetry prose drama fables article memoir letters reference lexicon).freeze
   
-  has_many :expressions, inverse_of: :work
+  has_many :expressions, inverse_of: :work, dependent: :destroy
   has_many :creations, dependent: :destroy
   has_many :persons, through: :creations, class_name: 'Person'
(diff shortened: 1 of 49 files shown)


app/models/work.rb
```diff
@@ -2,7 +2,7 @@
 class Work < ApplicationRecord
   GENRES = %w(poetry prose drama fables article memoir letters reference lexicon).freeze
   
-  has_and_belongs_to_many :expressions
+  has_many :expressions, inverse_of: :work
```
abartov, line 5: Don't we want dependent: :destroy on this relationship?  Deleting a work should mean deleting all its expressions, since we're not implementing compound expressions containing more than one work.

abartov (changes requested):
Looks good!  See the one little comment I made about dependent: :destroy, please.

app/models/work.rb
damisul: Good catch, thanks. Fixed

abartov (approved):

PR #162, opened 2022-06-16
Drop expressions_manifestation table

damisul:
Modified expressions <-> manifestations relation to be one-to-many instead of many-to-many.

Dropped expressions_manifestations table and added manifestations.expression_id column instead.

Updated all places where it was used and added specs to affected places.

Again, I did my best to update rake tasks, but I don't know how to test them

### DIFF
--- a/app/chewy/manifestations_index.rb
+++ b/app/chewy/manifestations_index.rb
@@ -2,7 +2,7 @@ class ManifestationsIndex < Chewy::Index
   settings 'index.max_result_window' => 40000 # Must be set to value greater than total number of works in db
 
   # works
-  define_type Manifestation.all_published.includes(expressions: :work) do
+  define_type Manifestation.all_published.includes(expression: :work) do
 #    field :title, analyzer: 'hebrew' # from https://github.com/synhershko/elasticsearch-analysis-hebrew
 #    field :fulltext, value: ->(manifestation) {manifestation.to_plaintext}, analyzer: 'hebrew'
     field :id, type: 'integer'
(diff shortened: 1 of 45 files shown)


abartov (approved):

abartov:
I have now deployed this in production, and the migration went smoothly. Good job!

However, there was a regression in the frbrize process: a NotNullException was being thrown (every time), because the null expression_id was preventing the new Manifestation from being saved before an existing Expression was associated with it. Changing the order of the save and the association resolved it. See my change in html_file.rb in this commit:
8bdfd7cbcc156ba88f4525287db6ac8f67085be2

But the fact this slipped through makes me think there is a problem of test coverage. How did it not fail in testing?

damisul:
Well... The key problem here is that we only recently started to add tests to project.
I'm adding specs to pieces of code I develop myself or change only. So we still have A LOT of places not covered with tests.

In this particular case I've did not found this problem in HTML Files controller, and I never worked with this controller previously, that's why this controller does not have spec yet.

Also I want to say, that even already added specs in many cases tests only basic logic, e.g. if call was succeed or not, but I can say, that even such basic specs helps a lot during massive refactorings like this one.

I saw projects where amount of testing code was greater than code of app itself, and I believe many of its tests were redundant. In my opinion it is important to keep balance between test coverage and sanity of tests.

abartov:
Yes, agreed on both counts!  And I am very grateful for all the tests you've added so far.  The lack of tests is Original Sin in this project, and I'm very happy we are now beginning to address it.  And yes, despite the partial coverage, we're already on much firmer grounds for refactoring, considering the significant user base and reliability demands of our production system.

(Eventually, I'll want to add functionality tests including the UI (using Web automation such as Selenium), but that can wait for a less busy time.)

PR #170, opened 2022-06-27
Optimized #translated_from_multiple_languages action.

damisul:
Optimized #translated_from_multiple_languages action.
On my machine it reduced page load time from 12 seconds to less than 2 seconds

- updated it to list only authors (old version displayed illustrators as well)
- added sorting of authors and works in lists

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -220,17 +220,23 @@ class AdminController < ApplicationController
 
   def translated_from_multiple_languages
     @authors = []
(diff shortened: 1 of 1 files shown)


damisul:
Additional question here: currently it also shows works in Hebrew: e..g if person has works in Russian and Hebrew, he will be displayed in this list, and all his Hebrew works will be displayed there together with russian works.

I'm not sure if this is an intended behavior.

abartov (approved):
Thanks. No, that's unintended behavior. But not urgent to fix.

PR #282, opened 2024-03-20
Works pagination fix

damisul:
I'm not sure why, but Kaminari's `total_pages` helper method was returning 100 when we had more than 100 pages on /works/all page.
That's why slider was limited to 100 pages only.

Most simple way to fix this was to calculate `total_pages` value without Kaminari.

Also this PR contains some adjustements to dockerfiles to match current production env (NOTE: we still need adjustements for sidekiq)


abartov (approved):

abartov:
Thank you! What odd behavior.

I have deployed this into production, as this was significant breakage.

PR #288, opened 2024-03-22
made email field mandatory in proof creation dialog.

damisul:
I've removed old code marked as legacy.
Also did some code cleanup (see comments below).

fixed bug in antispam check (proof created even if check failed)

### DIFF
--- a/app/controllers/proof_controller.rb
+++ b/app/controllers/proof_controller.rb
@@ -5,28 +5,25 @@ class ProofController < ApplicationController
 #  impressionist # log actions for pageview stats
 
   def create
-    if params['manifestation'].nil? # legacy site's buttons.js hack
-      unless params['what'].nil? or params['what'].empty? # don't bother capturing null submissions
-        #if is_blacklisted_ip(client_ip) # filter out spam
-        #  render plain: "OK"
-        #else
-          @p = Proof.new(:from => params['email'], :about => params['about'] || request.env["HTTP_REFERER"] || 'none', :what => params['what'], :subscribe => (params['subscribe'] == "yes" ? true : false), :status => 'new')
-          h = HtmlFile.find_by_url(@p.about.sub(/https?:\/\/.*benyehuda.org\//, ''))
-          @p.html_file = h unless h.nil?
-          @p.save!
-        #end
-      else
-        render plain: "OK"
-      end
-    else # new BYBE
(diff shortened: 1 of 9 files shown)


app/controllers/proof_controller.rb
```diff
@@ -5,28 +5,25 @@ class ProofController < ApplicationController
...
-        @p.save!
-      end
+    @errors = []
+    unless params[:ziburit] =~ /ביאליק/
+      @errors << I18n.t('proof.create.ziburit_failed')
     end
-    respond_to do |fmt|
-      fmt.html { }
-      fmt.js { flash[:notice] = I18n.t(:proof_thanks_html) }
+    if params[:from].blank?
```
damisul, line 12: Actually as far as I see Proof form says that email is optional, so perhaps we need to update wording there too.

app/views/proof/show.html.haml
```diff
@@ -28,7 +28,7 @@
     .graybox#highlight{style:'font-size: 14pt; border:1px;background-color:#dedede'}
       != sanitize(@p.highlight.gsub("\n","<br/>"))
   %br
-  %h3= t(:proof)+':'
+  %h3= t(:proof_heading)+':'
```
damisul, line 31: I've renamed 'proof' line in i18n file, as it conflicted with controller's name

config/locales/he.yml
```diff
@@ -444,7 +444,6 @@ he:
   activate_bit: הוסף תפקיד
   deactivate_bit: בטל תפקיד
   editor_bits: תפקידי עורך
-  proof_thanks_title: קיבלנו, תודה!
```
damisul, line 447: was used in proof html form I've removed

config/locales/he.yml
```diff
@@ -455,7 +454,6 @@ he:
   proofstatus_wontfix: לא שגוי
   proofstatus_escalated: טעון בדיקה נוספת
   proofstatus_assigned: מוקצה לטיפול
-  proofs: הגהות
```
damisul, line 458: was not used

app/controllers/proof_controller.rb
abartov: True! I am fixing the string and will commit after merging this.

abartov (approved):

PR #315, opened 2024-04-04
Corporate body crud

damisul:
Updated scaffold-generated controller to do CRUD for CorporateBody. It is located under http://localhost:3000/corporate_bodies

This PR introduce Simple_Form gem into a project, which is configured to use Bootstrap 4 (in future we can configure it to use different framework, or even some custom styles).

Please take a looks on shorter syntax of form code, and how we can manipulate with hints/placeholders/labels using i18n strings.

I've added i18n config for ActiveRecord and started filling it with data for CorporateBody. It will allow to use translated field names in all standard error messages.
Note, that Simple_form renders all validation errors below fields, and now they will be translated, e.g.:
изображение

Also in simple_form_he.yml you can override some titles for specific form, add hints, placeholder text, etc. (I've added few for demonstraiton, you can remove them).


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'http://rubygems.org'
 
-gem 'rails', '~> 6'
-gem 'rails-i18n' # , git: 'https://github.com/svenfuchs/rails-i18n.git' # , branch: 'rails-4-x' # For 4.x
+gem 'rails', '~> 6.1.7.7'
+gem 'rails-i18n', '~> 6.0' # version should match major version of Rails
```
damisul, line 4: It used version for Rails 7 and this is not recommended

app/controllers/corporate_bodies_controller.rb
```diff
@@ -1,106 +1,129 @@
+# frozen_string_literal: true
+
 class CorporateBodiesController < ApplicationController
-  before_action :set_corporate_body, only: %i[ show edit update destroy corp_toc]
+  before_action do |c| c.require_editor('edit_people') end
```
damisul, line 4: Perhaps we may want to add another bit to edit corporates, but for now I've used same check as for persons.

app/models/corporate_body.rb
```diff
@@ -13,6 +13,8 @@ class CorporateBody < ApplicationRecord
 
   has_one_attached :profile_image
 
+  validates_presence_of :name, :status
```
damisul, line 16: I believe at least those two field should be mandatory in model.

app/views/corporate_bodies/show.html.haml
```diff
@@ -1,36 +1,11 @@
...
+.btn-group
+  = link_to 'Edit', edit_corporate_body_path(@corporate_body), class: 'btn btn-default'
+  = link_to 'Back', corporate_bodies_path, class: 'btn btn-outline-secondary'
 
-= link_to 'Edit', edit_corporate_body_path(@corporate_body)
-\|
-= link_to 'Back', corporate_bodies_path
+- CorporateBodiesController::ATTRIBUTES_TO_SHOW.each do |attr|
+  %p
+    %b #{CorporateBody.human_attribute_name(attr)}:
```
damisul, line 10: This method takes attribute name from i18n files specific for given model (active_record.he.yml)

config/locales/active_record.en.yml
```diff
@@ -0,0 +1,24 @@
+en:
+  attributes:
```
damisul, line 2: Common attributes shared between different models

config/locales/simple_form.he.yml
```diff
@@ -0,0 +1,18 @@
...
+      text: 'required'
+      mark: '*'
+      # You can uncomment the line below if you need to overwrite the whole required html.
+      # When using html, text and mark won't be used.
+      # html: '<abbr title="required">*</abbr>'
+    error_notification:
+      default_message: "Please review the problems below:"
+    hints:
+      corporate_body:
+        name: Organization/Institution Name
```
damisul, line 15: This is a small gray hint below input control:
изображение

config/locales/simple_form.he.yml
```diff
@@ -0,0 +1,18 @@
...
+    placeholders:
+      corporate_body:
+        alternate_names: (if any)
```
damisul, line 18: This is the text shown in input when there is no content in it:
изображение

app/controllers/corporate_bodies_controller.rb
abartov: Yeah, they're the same bit.

abartov (approved):

PR #319, opened 2024-04-05
relaxed some linters

damisul:
This PRs relaxes/disables some linters, producing warnings which are not so critical but hard to fix (or we use different style in our project).


damisul:
Closing this will use those changes in different PR

PR #326, opened 2024-04-20
Refactor works mobile filter

damisul:
Refactored works/browse page to use same html form for desktop and mobile views.
Refactored several partials to share more code between authors and works browse pages.


app/views/manifestation/browse.html.haml
```diff
@@ -1,32 +1,28 @@
...
-        .by-card-v02.text-sorting#sort_filter_panel{style: 'display:none'}
-          .by-card-header-v02
-            .headline-2-v02= t(:sort_by)
-          .by-card-content-v02#filters_panel
-            .temporary
-      - else
-        .by-card-v02.text-sorting#sort_filter_panel{style: 'display:show'}
-          = render partial: 'browse_filters'
-    .col-12.col-lg-8
+      .by-card-v02.text-sorting#sort_filter_panel{ style: 'display:none' }
```
abartov, line 12: shouldn't this be conditional upon filters.empty? too?

damisul: Thanks, @abartov a good catch!
But we cannot decide if element should be shown or not on server side, as we need to know if this is a mobile view or desktop.
So I've added JS code to handle this.

Also added couple of other adjustements:
- browse_intro snippet should not be shown on mobile at all (to copy old behaviour)
- added scrolling to the top of the list when mobile filter is applied.

And modified authors browse page accordingly.

abartov (approved):

PR #343, opened 2024-05-13
Fixed HtmlFile.create_WEM_new method to work with new intellectual_property. Code cleanup.

damisul:
I've missed this during previous refactoring :(

Also there is a second commit, removing bunch of unused methods from Person and ApplicationController


damisul:
Included in https://github.com/abartov/bybeconv/pull/345

PR #345, opened 2024-05-14
Involved authorities

damisul:
This PR introduces new InvolvedAuthorities table to replace Creations and Realizers table.

It does not intoduces Authorities table yet, so InvolvedAuthority record can only refer to person.
Also I did not used polymorphic relation for item and added expression_id and work_id as separate columns instead. 
I believe it is better from performance point of view and more SQL-friendly approach.

Also this PR contains following changes not directly related to InvolvedAuthorities:
- It fixes HtmlFile.create_WEM_new method to work with new intellectual_property field (see logic for getting Expression.intellectual_property value from author and translator, I did my best from my understanding of logic)
- It removes bunch of unused methods and scopes from Person and ApplicationController classes


app/models/html_file.rb
```diff
@@ -632,43 +631,85 @@ def create_WEM_new(person_id, the_title, the_markdown, multiple, pub_status = ni
...
+
             q = (translator_id.nil? ? p : translator)
-            copyrighted = ((p.public_domain && q.public_domain) ? false : ((p.public_domain.nil? || q.public_domain.nil?) ? nil : true)) # if author is PD, expression is PD # TODO: make this depend on both work and expression author, for translations
-            e = Expression.new(title: tt, language: 'he', period: q.period, copyrighted: copyrighted, source_edition: publisher, date: year_published, comment: comments) # ISO codes
-            w.expressions << e
-            w.save!
-            c = Creation.new(work_id: w.id, person_id: p.id, role: :author)
-            c.save!
+            intellectual_property =
+              if p.intellectual_property_copyrighted? || q.intellectual_property_copyrighted?
```
damisul, line 647: I believe this is a correct way to determine intellectual_property type based on values from author and translator.

app/views/admin/missing_languages.html.haml
```diff
@@ -9,12 +9,10 @@
       %th
 
     - @mans.each do |m|
-      - xlator = m.expression.realizers.where(role: Realizer.roles[:translator])[0]
       %tr
         %td= link_to m.title, manifestation_show_path(m.id)
-        - person = m.expression.work.persons[0]
-        %td= person.nil? ? '' : person.name
-        %td= xlator.nil? ? '' : xlator.person.name
+        %td= m.authors.map(&:name).join(', ')
```
damisul, line 14: updated code to display all authors/translators

app/views/admin/suspicious_translations.html.haml
```diff
@@ -12,8 +12,8 @@
     - @mans.each do |m|
       %tr
         %td= link_to m.title, manifestation_show_path(m.id)
-        %td!= m.expression.work.persons.map{|p| link_to(p.name, authors_show_path(id: p.id))}.join(';')
```
damisul, line 15: I believe it was an error, as we need to check authors, not all work-level involved_authorities (previously it could include illustrators), so I've changed it to `authors`.

app/views/manifestation/_involved_authorities.html.haml
```diff
@@ -0,0 +1,17 @@
+%h3= t(:associated_persons)
```
damisul, line 1: I've created this partial to replaces old shared/associated_people partial to reduce code duplication

app/api/v1/entities/person.rb
```diff
@@ -23,19 +23,13 @@ class Person < Grape::Entity
...
-        end
-        expose :editor, documentation: { type: 'Integer', is_array: true } do |person|
-          Realizer.where(person_id: person.id).editor.joins(expression: :manifestations).pluck('manifestations.id').sort
-        end
-        expose :illustrator, documentation: { type: 'Integer', is_array: true } do |person|
-          Creation.where(person_id: person.id).illustrator.joins(work: { expressions: :manifestations }).pluck('manifestations.id').sort
+      expose :texts,
+             documentation: { desc: 'ID numbers of all texts this person is involved into with with role in each' },
+             if: ->(_person, options) { %w(texts enriched).include?(options[:detail]) } do
+        InvolvedAuthority.roles.each_key do |role|
```
damisul, line 29: Exposing all available roles including newly added

app/controllers/application_controller.rb
```diff
@@ -154,7 +150,7 @@ def randomize_works(how_many)
   def cached_authors_in_genre
     Rails.cache.fetch('au_by_genre', expires_in: 24.hours) do # memoize
       totals = Person.has_toc
-                     .joins(expressions: :work)
```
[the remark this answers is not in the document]
abartov: Yes, you are correct!

app/models/expression.rb
```diff
@@ -25,12 +26,8 @@ class Expression < ApplicationRecord
 
   validates :intellectual_property, presence: true
 
-  def editors
```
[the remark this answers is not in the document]
abartov: I don't understand: why can't we return the editors associated with the Expression here?

app/models/person.rb
```diff
@@ -222,25 +235,35 @@ def all_languages
   end
 
   def original_works
-    Manifestation.all_published.joins(expression: [work: :creations]).where("creations.person_id = #{self.id}")
+    Manifestation.all_published
+                 .joins(expression: [work: :involved_authorities])
+                 .where(involved_authorities: { person_id: id })
```
abartov, line 240: wouldn't this include translations, too? This is supposed to return only Manifestations where this person is NOT the translator.

app/models/person.rb
```diff
@@ -222,25 +235,35 @@ def all_languages
...
+                         .merge(InvolvedAuthority.where(person_id: id))
+    xlats = Manifestation.joins(expression: :involved_authorities)
+                         .merge(InvolvedAuthority.role_translator.where(person_id: id))
+    (works + xlats).uniq.sort_by(&:sort_title)
   end
 
   def original_work_count_including_unpublished
-    Manifestation.joins(expression: { work: :creations }).merge(Creation.where(person_id: self.id)).count
+    Manifestation.joins(expression: { work: :involved_authorities })
+                 .merge(InvolvedAuthority.where(person_id: id)).count
```
abartov, line 260: likewise.

app/models/person.rb
```diff
@@ -263,7 +286,10 @@ def title # convenience method for polymorphic handling (e.g. Taggable)
   def original_works_by_genre
     ret = {}
     get_genres.map{|g| ret[g] = []}
-    Manifestation.all_published.joins(expression: [work: :creations]).includes(:expression).where("creations.person_id = #{self.id}").each do |m|
+    Manifestation.all_published
+                 .joins(expression: [work: :involved_authorities])
+                 .includes(:expression)
+                 .merge(InvolvedAuthority.where(person_id: id)).each do |m|
```
abartov, line 292: likewise.

app/models/person.rb
```diff
@@ -285,12 +313,18 @@ def featured_work
   end
 
   def latest_stuff
-    latest_original_works = Manifestation.all_published.left_joins(expression: [:realizers, work: :creations]).includes(:expression).where(creations: {person_id: self.id}).order(created_at: :desc).limit(20)
+    latest_original_works = Manifestation.all_published
+                                         .joins(expression: [work: :involved_authorities])
+                                         .where(involved_authorities: { person_id: id })
+                                         .order(created_at: :desc).limit(20)
```
abartov, line 319: likewise.

app/services/random_author.rb
```diff
@@ -8,12 +8,12 @@ def call(genre = nil)
                        <<~SQL.squish,
                          exists (
                            select 1 from
-                             creations c
-                             join works w on (c.work_id = w.id)
+                             involved_authorities ia
```
abartov, line 11: again, don't we lose the distinction here, if we don't filter by role?  This could now return someone who only *translated* a work in a particular genre.

app/models/expression.rb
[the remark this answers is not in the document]
damisul: Well... Probably I used wrong wording.

I wrote this comment while thinking about: https://github.com/abartov/bybeconv/blob/5f265048d8286dfc8fef6cf429feb7d98f6ab50a/app/controllers/manifestation_controller.rb#L845C1-L847C1
I believe in that case we should fetch illustrators and editors both from Work and Expression. So I've added `Manifestation.involved_authorities_by_role` method for that puropse which merges values from Work and Expression.

And there is an `involved_authorities_by_role` method in `Work` and `Expression` too if we need value from specific level.

app/models/person.rb
damisul: Well... I believe this will not include translators, because translator can only be specified on Expression level (we have validation for this). But this will include all works where Person has any role possible on Work level (e.g. illustrator).

I wonder if we should limit it to return only works where person is an author? In this case I'd propose to remove this method at all and use new `Person.published_manifestations(:author)` method instead. @abartov  WDYT?

app/services/random_author.rb
damisul: Yes. you're right. Fixed. 

If we need to get only authors here we need to add filter by role. By the way, old code, also could return illustrators. I believe it was mitigated by the fact that only person with TOC were selected, but I've added explicit filter here.

abartov (approved):

PR #353, opened 2024-05-16
Refactor wikidata_id column to wikidata_uri

damisul:
See https://github.com/abartov/bybeconv/issues/352

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -1,5 +1,6 @@
 include BybeUtils
 class Person < ApplicationRecord
+  WIKIDATA_URI_PATTERN = %r{\Ahttps://wikidata.org/wiki/q[0-9]+\z}
 
   enum gender: %i(male female other unknown)
   enum period: %i(ancient medieval enlightenment revival modern)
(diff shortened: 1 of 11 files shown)


config/locales/active_record.he.yml
```diff
@@ -18,3 +18,4 @@ he:
         intellectual_property: מצב זכויות יוצרים
       person:
         intellectual_property: מצב זכויות יוצרים
+        wikidata_uri: Wikidata URI
```
damisul, line 21: Need a translation

app/models/person.rb
```diff
@@ -48,11 +49,16 @@ class Person < ApplicationRecord
...
   # validations
   validates :name, :intellectual_property, presence: true
+  validates :wikidata_uri, format: WIKIDATA_URI_PATTERN, allow_nil: true
 
   validates_attachment_content_type :profile_image, content_type: /\Aimage\/.*\z/
 
   update_index('people'){self} # update PeopleIndex when entity is updated
 
+  before_validation do
+    self.wikidata_uri = wikidata_uri.blank? ? nil : wikidata_uri.strip.downcase
```
abartov, line 59: the Q in https://wikidata.org/wiki/Q1234 is uppercase, canonically. It's true there is an auto-redirect in place at wikidata.org, but it's still incorrect for us to store a lowercase q. Please fix this.

abartov (changes requested):
see my requested change. Also, the tests don't seem to pass now.

app/models/person.rb
damisul: Oh, thanks, did not know that.
The problem why specs failing is because I've added wikidata_uri.downcase call in Person.before_validation block.
Just wanted to make regexp validation simpler with 'normalizing' input.

But as URIs are case-sensitive I've added more complex normalization logic to make Q uoppercase while rest of URI is downcased.

abartov (approved):

PR #358, opened 2024-05-20
Removed dates and affiliation columns from People table.

damisul:
As we discussed in Telegram

### DIFF
--- a/app/views/admin/missing_copyright.html.haml
+++ b/app/views/admin/missing_copyright.html.haml
@@ -6,13 +6,11 @@
     %tr
       %th= t(:name)
       %th= t(:life_years)
-      %th= t(:deathdate)
       %th
     - @authors.each do |au|
       %tr
         %td= au.name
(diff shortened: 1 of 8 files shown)


app/views/admin/missing_copyright.html.haml
```diff
@@ -6,13 +6,11 @@
     %tr
       %th= t(:name)
       %th= t(:life_years)
-      %th= t(:deathdate)
```
damisul, line 9: I've also removed separate deathdate column, as deathdate is displayed in life_years

spec/controllers/admin_controller_spec.rb
```diff
@@ -42,18 +42,6 @@
     it { is_expected.to be_successful }
   end
 
-  describe '#suspicious_titles' do
```
damisul, line 45: Removed duplicated test

abartov (approved):

PR #363, opened 2024-05-30
Authorities

damisul:
This PR splits existing People table to Authorities (storing common attributes) and People (storing person-only attributes) tables. It also adds new CorporateBodies table to store organization-specific properties.

I've modified existing tables to point to Authorities table insted of People and updated related code.

Same form is used for CRUD both for person- and corporate_body- authorities (we hide/show related fields depending on specified type).

Also I've renamed PeopleIndex to AuthoritiesIndex, now it stores data both for People and CorporateBodies.

NOTE: I've did not touched PeopleAPI yet, only updated it to work with new structure. It will ignore CorporateBodies for now. I think we should rename endpoint to Authorities later (let's create separate ticket for this).

Also I did my best to update bib_controller. but I have little knowledge of its logic. So I'd propose to test it deeply, as there are more chances for bugs.

And as usual there are number of i18n strings waiting for translation to Hebrew.

And one more smaller issue to be fixed, is to rename url_helper redirecting to TOC of authority. I kept it 'person_path', but in reality it leads to 'toc' page. I'd propose to rename it to 'authority_toc' instead. WDYT? I can do it in a separate PR later.


app/api/v1/entities/person.rb
```diff
@@ -2,17 +2,28 @@ module V1
...
     class Person < Grape::Entity
       expose :id, documentation: { type: :Integer }
-      expose :url, documentation: { desc: "Canonical URL of the person at Project Ben-Yehuda (useful for giving credit and allowing users to click through)" } do |person|
-        Rails.application.routes.url_helpers.bib_person_url(person)
+      expose :url, documentation: {
+        desc: 'Canonical URL of the person at Project Ben-Yehuda (useful for giving credit and allowing' \
+              'users to click through)'
+      } do |au|
+        Rails.application.routes.url_helpers.bib_authority_url(au)
       end
```
abartov, line 10: Wait, I've only now noticed this URL never pointed to the right place, to begin with: the canonical URL for a person (and now for an authority) should be /author/\d+, i.e. the public table-of-contents view for that person. The bib_person_url is internal, for the few users authorized to use the bibliographical workshop in the back-end.  I missed this while previously reviewing the API, but this field should return authority_url, not bib_authority_url.

abartov (approved):

PR #368, opened 2024-06-04
Code cleanup

damisul:
Removed method `AuthorsController.all` and `all.html.haml` view as it is not mapped in routes.rb
Removed unused partial `more_from_author`, as far as I see it is not used too


app/controllers/authors_controller.rb
```diff
@@ -259,10 +259,6 @@ def browse
     end
   end
 
-  def all
```
abartov, line 262: This used to lead to /authors/all.html.haml, which you are right to delete, but the action is still needed, and is referred to in config/routes, because the URL /author/all did exist at one time, and may be linked to from outside our site, and we want not to break it, but to seamlessly redirect it to just /authors.
(This could also be accomplished in the Web server through URL rewriting for a tiny performance gain, but I prefer to keep the Web server config as trivial as possible.)

damisul: Oops, yes, there is a declaration in routes.rb for it. I simply overlooked it.
Returned it back

abartov (approved):

PR #369, opened 2024-06-04
Authorities path refactoring and bugfixing

damisul:
This PR renames person_path helper to authority_path.

While working on this found and fixed several leftovers not fixed during authorities refactoring:
- fixed several methods in  HtmlFile controller
- updated tags management code to link to Authorities instead of People
- updated SearchController to use AuthoritiesIndex instead of PeopleIndex
added spec for found places.


abartov (approved):
Great work! 
And thank you for writing those controllers specs I never wrote. (My rspec fu is still quite weak.)

PR #374, opened 2024-06-12
Authors bugfixes

damisul:
Fixed several bugs introduced during Authorities refactoring:
- fixed filtering by person's traits (gender, perdiod, birth and death years)
- updated authors#to_manual_toc, #create_toc and #destroy actions to work with Authorities instead of People. There are many changes in #to_manual_toc, but they should not affect logic, I did them to satisfy linters

P.S. also please see https://github.com/abartov/bybeconv/issues/373

### DIFF
--- a/app/controllers/authors_controller.rb
+++ b/app/controllers/authors_controller.rb
@@ -12,7 +12,7 @@ class AuthorsController < ApplicationController
 
   before_action :set_author,
                 only: %i(show edit update destroy toc edit_toc prep_for_print print all_links delete_photo
-                         whatsnew_popup latest_popup publish)
+                         whatsnew_popup latest_popup publish create_toc to_manual_toc)
   autocomplete :tag, :name, :limit => 2
 
   def publish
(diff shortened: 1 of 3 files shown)


config/locales/he.yml
```diff
@@ -1109,6 +1109,7 @@ he:
   no_such_item: אין פריט כזה.
   item: פריט
   created_toc: נוצר דף יוצר עם תבנית ריקה
+  already_has_toc: Author already has TOC
```
damisul, line 1112: When wrote spec for #to_manual_toc it turned out we miss this string

app/controllers/authors_controller.rb
```diff
@@ -282,81 +282,63 @@ def by_tag
   end
 
   def create_toc
-    author = Person.find(params[:id])
-    unless author.nil?
```
damisul, line 286: Actually we have many places like this all over the code. They are useless, as ActiveRecord.find method raises an RecordNotFound exception if record with given id is not found. So we never get author.nil? condition satisfied,  but code expects nil to be returned in such case.

abartov (approved):

PR #379, opened 2024-06-14
Impressions and YearTotals update

damisul:
added migration to update impressions and year_totals records to reference authorities instead of people

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2024_06_09_165640) do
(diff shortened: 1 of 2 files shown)


db/schema.rb
abartov, line 1: This seems to have slipped in by mistake.

abartov (changes requested):
exclude db/schema.rb with ingestible

db/schema.rb
damisul: Hm... actually not. I want to merge this into a master branch and master branch should not know anything about ingestible table for now

abartov: Oh, sorry! I instinctively misread the diff the other way around! :)
*I* am the one who mistakenly committed the ingestible. Sorry!

abartov (approved):

PR #381, opened 2024-06-14
Render google-analytics script only in production environment

damisul:
This PR disables rendering of google analytics script in development environment


damisul:
It will speedup page loading in dev environment a bit and will not count development actions in stats

PR #382, opened 2024-06-14
Updated Search to show Authorities first

damisul:
Updated multiindex search to explicitely sort items to sort items in following order Authorities, Manifestations, Dicts using script sorting.

By default Kaminary sorts by index name in **desc** order, so after Person to Authority refactoring we got [Manifestations, Dicts, Authorities]. Previously (before index renaming) it was [People, Manifestations, Dicts].

Also I fixed another minor issue: it always displayed items with index starting from 1 even if non-first page of results was requested. Now second page will start from 26, third from 51, etc.

### DIFF
--- a/app/lib/manifestations_search.rb
+++ b/app/lib/manifestations_search.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'multi_index_search_request'
 class ManifestationsSearch
   include ActiveData::Model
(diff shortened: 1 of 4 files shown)


app/lib/manifestations_search.rb
```diff
@@ -29,34 +31,53 @@ def search
...
+          :name,
+          :other_designation,
+          :author_string,
+          :fulltext,
+          :deftext
+        ],
+        query: query,
+        default_operator: 'and'
+      }
+    ).order('_index' => { order: :asc })
```
damisul, line 52: main change in this PR.

app/lib/manifestations_search.rb
```diff
@@ -29,34 +31,53 @@ def search
...
+    index.filter(term: { genre: genre }) if genre?
   end
 
   # For filtering on years, we will use range filter.
   # Returns nil if both min_year and max_year are not passed to the model.
   def orig_pub_year_filter
-    body = {}.tap do |body|
-      body.merge!(gte: min_orig_pub_year) if min_orig_pub_year?
-      body.merge!(lte: max_orig_pub_year) if max_orig_pub_year?
+    body = {}.tap do |new_body|
```
damisul, line 63: avoided external var shadowing

abartov (changes requested):
We do actually want DictIndex to be last for now (because works are much more frequently the results people want than dictionary entries). At least until we implement the tabbed results display, let's enforce that.

Perhaps we can keep the search code simple by renaming DictIndex to ZDictIndex or something? Should be possible using Chewy::Index's base_name or index_name, I think.

abartov (approved):

PR #384, opened 2024-06-14
Made it possible to switch UI to english in development

damisul:
If BY_LANGUAGE env variable is provided it will try to use given locale as default.
For now only en or he locale is accepted (and only in dev environment)

I've added dotenv gem to simply manage env variables in development and test. I plan to use it also in https://github.com/abartov/bybeconv/issues/166

You can create `.env.development.local` file and override BY_LANGUAGE value there instead of overwriting `.env.development`

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,0 +1,4 @@
+# Common environment variables for development environment.
+# Usually you don't need to change this file directly, if you need to override some vars with values specific to
+# your machine, create an `.env.development.local` file and set values there.
+BY_LANGUAGE=he
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -27,7 +27,8 @@ Rails.application.configure do
     config.cache_store = :null_store
   end
 
-  config.i18n.enforce_available_locales = false
+  # In development we allow using of different locales (en for now) for non-hebrew speaking developers
+  config.i18n.default_locale = ENV['BY_LANGUAGE']&.to_sym || :he
 
   # Don't care if the mailer can't send
   config.action_mailer.raise_delivery_errors = true
@@ -54,7 +55,6 @@ Rails.application.configure do
   #config.assets.debug = true
   config.eager_load = false
   # config.public_file_server.enabled = true # Rails 5.x?
-  config.i18n.available_locales = :he
 
   # Store Active Storage files locally.
   config.active_storage.service = :local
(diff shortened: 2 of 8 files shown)


.env.development
```diff
@@ -0,0 +1 @@
+BY_LANGUAGE=he
```
damisul, line 1: provided as an example for dotenv config. To override it locally create a new file .env.development.local (should not be committed to git)

config/environments/production.rb
```diff
@@ -68,16 +68,11 @@
   # Enable threaded mode
   # config.threadsafe!
 
-  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
-  # the I18n.default_locale when a translation can not be found)
-  config.i18n.fallbacks = true
```
damisul, line 73: already specified in application.rb

config/environments/production.rb
```diff
@@ -68,16 +68,11 @@
...
-
   # Send deprecation notices to registered listeners
   config.active_support.deprecation = :notify
 
   config.action_mailer.delivery_method = :sendmail # let's start sending mail :)
   config.eager_load = true
-  config.i18n.available_locales = :he
```
damisul, line 80: already specified in application.rb

config/environments/production.rb
```diff
@@ -102,10 +97,6 @@
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
   # config.action_mailer.raise_delivery_errors = false
 
-  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
-  # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
```
damisul, line 107: duplicate line

config/environments/development.rb
```diff
@@ -54,7 +55,6 @@
   #config.assets.debug = true
   config.eager_load = false
   # config.public_file_server.enabled = true # Rails 5.x?
-  config.i18n.available_locales = :he
```
damisul, line 57: defined in application.rb

config/locales/en.yml
```diff
@@ -1,4 +1,5 @@
 en:
+  edit_metadata: Edit metadata
```
damisul, line 2: added for tests

abartov (approved):

PR #386, opened 2024-06-19
Implemented locks mechanics for ingestibles

damisul:
Added two new columns to Ingestibles table: locked_by_user_id and locked_at.

On every action which may changed ingestible, it now tries to lock ingestible.
Record considered locked if locked_at is not more than 15 minutes ago.

Also fixed several i18n-related issues (wrong scope for i18n string ingestible -> ingestibles (controller name)).

Added specs for IngestiblesController.


app/assets/stylesheets/application.css
```diff
@@ -400,11 +400,11 @@ body.modal-open{
 .modal{
   padding-right: 0 !important;
 }
-.error, .notice, .success {padding:.8em;margin-bottom:1em;border:2px solid #ddd;}
```
damisul, line 403: Rails defines two types of flashes out of the box: notice and alert, and they have fancy accessors like `flash.alert` instead of `flash[:alert]`.

I've added styles for alert to be displayed in same way as error. Actually I'd propose to use `alert` everywhere instead of `error` to follow Rails defaults.

app/controllers/concerns/lock_ingestible_concern.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Logic to lock Ingestible object, to be included in related controllers
+module LockIngestibleConcern
```
damisul, line 4: I've moved this method to separate concern because I expect some other controllers to work with ingestibles in future

app/controllers/concerns/lock_ingestible_concern.rb
```diff
@@ -0,0 +1,13 @@
...
+  extend ActiveSupport::Concern
+
+  def try_to_lock_ingestible
+    return if @ingestible.obtain_lock(current_user)
+
+    redirect_to ingestibles_path,
+                alert: t('.ingestible_locked', user: @ingestible.locked_by_user.name)
```
damisul, line 11: Initially I wanted to redirect to show page, but currently show page renders edit form, so I've redirected to ingestibles list instead

app/controllers/ingestibles_controller.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class IngestiblesController < ApplicationController
+  include LockIngestibleConcern
+
+  before_action { |c| c.require_editor('edit_catalog') }
```
damisul, line 6: Added access check. We need user to be authorized to be able to lock ingestible

app/controllers/ingestibles_controller.rb
```diff
@@ -1,7 +1,11 @@
...
   before_action :set_ingestible, only: %i(show edit update destroy addauth rmauth)
+  before_action :try_to_lock_ingestible, only: %i(show edit update destroy addauth rmauth)
```
damisul, line 8: If we'll have show page later we should remove locking on show page

app/controllers/ingestibles_controller.rb
```diff
@@ -71,25 +71,18 @@ def rmauth
...
-      if @ingestible.update(ingestible_params)
-        format.html { redirect_to edit_ingestible_url(@ingestible), notice: 'Ingestible was successfully updated.' }
-        format.json { render :edit, status: :ok, location: @ingestible }
-      else
-        format.html { render :edit, status: :unprocessable_entity }
-        format.json { render json: @ingestible.errors, status: :unprocessable_entity }
-      end
+    if @ingestible.update(ingestible_params)
+      flash.now.notice = t('.success')
+      render :edit
```
damisul, line 76: used `render :edit` instead of `redirect to edit`, should be a little bit faster

app/models/ingestible.rb
```diff
@@ -4,19 +4,25 @@
 
 # Ingestible is a set of text being prepared for inclusion into a main database
 class Ingestible < ApplicationRecord
+  LOCK_TIMEOUT_IN_SECONDS = 900
```
damisul, line 7: 15 minutes

app/models/ingestible.rb
```diff
@@ -158,4 +164,19 @@ def update_buffers
...
+  end
+
+  def obtain_lock(user)
+    return false if locked? && locked_by_user_id != user.id
+
+    update!(locked_at: Time.zone.now, locked_by_user: user)
+    return true
+  end
+
+  def release_lock
```
damisul, line 179: I've added this method, but actually I don't found proper place to use it.
So for now all locks will stay active for 15 minutes after last access

config/locales/en.yml
```diff
@@ -13,8 +13,21 @@ en:
         db_intellectual_property: Intellectual property in DB
         has_non_public_domain_authority: Has non public domain authority
   ingestibles:
```
damisul, line 15: This is controller-related i18n resources, so to use shorthand version e.g. `.success` instead of `ingestible.update.success` we should prefix with controller name (ingestible**s** instead of ingestible)

app/controllers/concerns/lock_ingestible_concern.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Logic to lock Ingestible object, to be included in related controllers
+module LockIngestibleConcern
+  extend ActiveSupport::Concern
+
+  def try_to_lock_ingestible
+    return if @ingestible.obtain_lock(current_user)
```
damisul, line 8: This method will prolong lock (for next 15 minutes), if user already had lock on ingestible

app/assets/stylesheets/application.css
abartov: thank you. I think these are remnants of ancient Rails habits (maybe 2.x?), when there used to be a flash[:error].

app/controllers/concerns/lock_ingestible_concern.rb
abartov: Yes. There's basically no scenario where a back-end editor user would want to just view an ingestible. They are interim entities that you work with until they are ready, and then they feed into the main catalog and are archived forever.  

Though I guess we may occasionally want to allow another editor to look at an ingestible (e.g. an admin or a developer to help another editor understand some error or bug), so perhaps a view does make sense to implement later, but I didn't want to spend any time at all on it before we have the designed screens anyway.

app/models/ingestible.rb
abartov: (clearer if written as 15*60 :))

app/models/ingestible.rb
abartov: Yeah, maybe we would implement an admin-override button that would force a lock open. Good planning ahead.

app/models/ingestible.rb
damisul: Ok, fixed!

abartov (approved):

PR #389, opened 2024-06-20
Ingestible authorities refactor

damisul:
**This PR should be merged after https://github.com/abartov/bybeconv/pull/386 (it includes all changes from 386)**

This PR moves addauth and rmauth methods from IngestiblesController to new IngestibleAuthoritiesController.
Declared it as nested resources inside ingestibles resources


app/controllers/ingestibles_controller.rb
```diff
@@ -4,8 +4,8 @@ class IngestiblesController < ApplicationController
   include LockIngestibleConcern
 
   before_action { |c| c.require_editor('edit_catalog') }
```
abartov, line 6: this is missing in the new separate controller

abartov (changes requested):
missing access check in new controller.

app/controllers/ingestibles_controller.rb
damisul: Oops, thank you, overlooked this. Fixed now

PR #392, opened 2024-06-24
bib_done radio buttons on authorities form was not displayed properly

damisul:
In simple form we need to use true/false instead of 0/1

Also added some common i18n strings to english locale.

Also fixed one issue not related to this issue: on featured_authors page title from featured content page was used.


app/views/admin/featured_author_list.html.haml
```diff
@@ -1,5 +1,5 @@
 .backend
-  %h2= t(:featured_content_management)
```
damisul, line 2: wrong heading was used

abartov (approved):

PR #401, opened 2024-07-14
Ingestible paging

damisul:
Added UI to page between individual texts inside an ingestible.
For now only markdown and title can be edited (I expect we'll have an authorities too?)

One of limitations of current implementation is that when saving changes to ingestible text we will loose any changes to other ingestible metadata (if any was changed) will be lost.

we can either add a warning message about this, or consider splitting metadata and markdown update into a separate actions (as we currently have in manifestations)

### DIFF
--- a/app/controllers/ingestible_texts_controller.rb
+++ b/app/controllers/ingestible_texts_controller.rb
@@ -4,6 +4,7 @@
 class IngestibleTextsController < ApplicationController
   include LockIngestibleConcern
 
+  before_action { |c| c.require_editor('edit_catalog') }
   before_action :set_ingestible
   before_action :try_to_lock_ingestible
 
--- a/app/controllers/ingestible_texts_controller.rb
+++ b/app/controllers/ingestible_texts_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# Controller to work with individual texts within an Ingestible object
+class IngestibleTextsController < ApplicationController
+  include LockIngestibleConcern
+
+  before_action :set_ingestible
+  before_action :try_to_lock_ingestible
+
+  def edit; end
+
+  def update
+    @ingestible.texts[@text_index] = IngestibleText.new(params.require(:ingestible_text).permit(:content, :title))
+    @ingestible.save!
+    redirect_to edit_ingestible_path(@ingestible, text_index: @text_index), notice: t(:updated_successfully)
+  end
+
+  private
+
+  def set_ingestible
+    @ingestible = Ingestible.find(params[:ingestible_id])
+    # Id is a zero-based index of text inside of ingestible texts collection
+    @text_index = params[:id].to_i
+  end
+end
--- a/app/controllers/ingestibles_controller.rb
+++ b/app/controllers/ingestibles_controller.rb
@@ -4,8 +4,8 @@ class IngestiblesController < ApplicationController
   include LockIngestibleConcern
 
   before_action { |c| c.require_editor('edit_catalog') }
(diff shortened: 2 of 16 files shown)


app/controllers/ingestibles_controller.rb
```diff
@@ -28,11 +28,11 @@ def new
   # GET /ingestibles/1/review
   def review
   end
-  
+
   # GET /ingestibles/1/edit
   def edit
     @ingestible.update_parsing # refresh markdown or text buffers if necessary
-    @html = MultiMarkdown.new(@ingestible.markdown).to_html.force_encoding('UTF-8').gsub(/<figcaption>.*?<\/figcaption>/,'') # remove MMD's automatic figcaptions
+    @html = MarkdownToHtml.call(@ingestible.markdown)
```
damisul, line 35: Extracted this code into a service. It would be good to update all places where we use this to use this service instead of copy/pasting code

app/models/ingestible.rb
```diff
@@ -165,6 +171,11 @@ def update_buffers
     save
   end
 
+  def texts
```
damisul, line 174: Added this helper method to work with IngestibleText objects instead of hashes

app/models/ingestible.rb
```diff
@@ -25,6 +25,12 @@ class Ingestible < ApplicationRecord
 
   # after_commit :update_parsing # this results in ActiveStorage::FileNotFoundError in dev/local storage
 
+  before_save do
+    if @texts.present?
```
damisul, line 29: if we modified parsed texts array we serialize it back to JSON before saving ingestible

app/views/ingestibles/_form.html.haml
```diff
@@ -1,22 +1,23 @@
-= form_for @ingestible do |f|
```
damisul, line 1: It is hard to use single form approach for ingestible edit page, as we need to update different entities from it.
So I've made every tab to have own form updated separately. This approach has some drawbacks though (see PR description)

app/controllers/ingestibles_controller.rb
```diff
@@ -50,13 +50,18 @@ def create
...
     if @ingestible.update(ingestible_params)
-      flash.now.notice = t('.success')
-      render :edit
+      redirect_to edit_ingestible_url(@ingestible), notice: t('.success')
     else
       render :edit, status: :unprocessable_entity
     end
   end
 
+  def update_markdown
```
damisul, line 59: Added separate action to update markdown.
We can move markdown parsing code here instead of calling it on every edit request

app/models/ingestible.rb
```diff
@@ -0,0 +1,193 @@
...
+      raise 'Conversion error'
+    ensure
+      tmpfile.close
+    end
+  end
+
+  # copied from HtmlFileController's new_postprocess
+  def postprocess(buf)
+    # join lines in <span> elements that, er, span more than one line
+    buf.gsub!(%r{<span.*?>.*?\n.*?</span>}) { |thematch| thematch.sub("\n", ' ') }
```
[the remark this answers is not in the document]
damisul: I did not touched this code. So it seems we already had this problem before

app/controllers/ingestible_texts_controller.rb
abartov, line 1: missing access check?

```
before_action { |c| c.require_editor('edit_catalog') }
```

damisul: Thank you! Fixed!

abartov (approved):

PR #403, opened 2024-07-22
Catch all collection

damisul:
One of prerequisites of creating root collection functionality is to implement 'catch all' colleciton for all uncollected works per authority.

This PR adds new collection type 'uncollected', and adds authorities.uncollected_works_collection_id column to link those collections with authorities. So each authority can have own collection.

There is a new service RefreshUncollectedWorksCollection, used to create or update uncollected collection per authority.

For test purpose I've added a new link 'Refresh uncollected works collection' to Author's TOC page, which calls this service. Later we can remove it and made service to be run automatically on certain events. This service does two things:
- it removes from uncollected works collection all items included in any collection of other type
- it adds works which are either not included in any collections, or only added to other 'uncollected' collections to other authorities.

NOTE: I've created new AuthoritiesController in admin namespace for this purpose. I propose to move all non-public actions related to authorities theres instead of AuthorsController (in future we can make AuthorsController to be responsible for public functionality, and admin/authorities controller to be responsible for editor functions).

### DIFF
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -37,6 +37,8 @@ class Authority < ApplicationRecord
 
   belongs_to :person, optional: true
   belongs_to :corporate_body, optional: true
+  belongs_to :root_collection, class_name: 'Collection', optional: true
+  belongs_to :uncollected_works_collection, class_name: 'Collection', optional: true
 
   attr_readonly :person, :corporate_body # Should not be modified after creation
 
@@ -74,6 +76,8 @@ class Authority < ApplicationRecord
   # validations
   validates :name, :intellectual_property, presence: true
   validates :wikidata_uri, format: WIKIDATA_URI_PATTERN, allow_nil: true
+  validates :uncollected_works_collection, uniqueness: true, allow_nil: true
+  validate :validate_collection_types
   validate :validate_linked_authority
 
   validates_attachment_content_type :profile_image, content_type: %r{\Aimage/.*\z}
@@ -334,11 +338,8 @@ class Authority < ApplicationRecord
     publish! if awaiting_first?
   end
 
-  def root_collection
-    return @root_collection if @root_collection
-    return @root_collection = Collection.find(root_collection_id) if root_collection_id
-
-    return @root_collection = generate_root_collection!
+  def obtain_root_collection
+    self.root_collection ||= generate_root_collection!
   end
 
   def generate_root_collection!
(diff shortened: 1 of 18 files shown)


app/models/authority.rb
```diff
@@ -334,11 +338,8 @@ def publish_if_first!
     publish! if awaiting_first?
   end
 
-  def root_collection
-    return @root_collection if @root_collection
-    return @root_collection = Collection.find(root_collection_id) if root_collection_id
-
-    return @root_collection = generate_root_collection!
+  def obtain_root_collection
```
damisul, line 341: root_collection method was very misleading, as it hid actual database relation. So I've added belongs_to root_collection, and renamed this method to not hide relation.

app/models/authority.rb
```diff
@@ -401,4 +402,16 @@ def validate_linked_authority
     errors.add(:base, :no_linked_authority) if person.nil? && corporate_body.nil?
     errors.add(:base, :multiple_linked_authorities) if person.present? && corporate_body.present?
   end
+
+  # rubocop:disable Style/GuardClause
+  def validate_collection_types
```
damisul, line 407: Additional validation: root_collection must be of 'root' type and uncollected_works_collection must be of 'uncollected' type.

app/models/collection.rb
```diff
@@ -34,7 +34,16 @@ class Collection < ApplicationRecord
...
-  enum collection_type: { volume: 0, periodical: 1, periodical_issue: 2, series: 3, root: 4, other: 5 }
+  # uncollected is used to group authority's works not belonging to any other collections. Each authority can have one.
+  enum collection_type: {
+    volume: 0,
+    periodical: 1,
+    periodical_issue: 2,
+    series: 3,
+    root: 4,
+    other: 5,
+    uncollected: 100
```
damisul, line 45: new collection type.

app/views/authors/toc.html.haml
```diff
@@ -60,7 +60,13 @@
                   != "&nbsp;&nbsp;&nbsp;"
                   %span.static-btn
                     %b= link_to t(:edit_metadata), authors_edit_path(id: @author.id)
-        
+
+                  != '&nbsp;&nbsp;&nbsp;'
+                  %span.static-btn
+                    %b= link_to 'Refresh uncollected works',
```
damisul, line 66: Temporary link to call service to refresh uncollected works collection. I did not bothered about I18n, as it is a temporary solution.

app/views/shared/_collection.html.haml
```diff
@@ -37,5 +37,5 @@
               - else
                 .headline-3-v02
                   = t(:text)
-                  :
+                  \:
```
damisul, line 40: Without escaping this line caused error.

spec/models/tag_spec.rb
```diff
@@ -96,11 +96,11 @@
 
   describe 'taggings scopes' do
     let(:first) do
-      described_class.create!(name: Faker::Science.science, creator: create(:user), status: 'approved')
+      described_class.create!(name: Faker::Science.unique.science, creator: create(:user), status: 'approved')
```
damisul, line 99: This is unrelated to PR, but I faced spec error. It seems, that sometimes Faker produced two tags with same name, and it caused spec to fail. Added unique instruction, to make Faker to generate unique values.

spec/models/tag_spec.rb
```diff
@@ -96,11 +96,11 @@
 
   describe 'taggings scopes' do
     let(:first) do
-      described_class.create!(name: Faker::Science.science, creator: create(:user), status: 'approved')
+      described_class.create!(name: Faker::Science.unique.science, creator: create(:user), status: :approved)
```
damisul, line 99: not related to PR directly, but sometimes those specs failed if Faker generated two tags with same name.

abartov (approved):

PR #404, opened 2024-07-24
Fixed tag_spec

damisul:
In some cases this spec failed due to non-unique tag name. Used Faker's unique feature to prevent this.


abartov (approved):

abartov:
Indeed, thank you. I think I fixed that on the collections branch, but not in master yet.

PR #405, opened 2024-07-27
added validation check to prevent creation of cyclic collections

damisul:
While building TOC it is important to not have cyclic dependencies between collections.
This PR adds validation check to CollectionItem which will not allow to create a CollectionItem if it produces cyclic collections dependency.


damisul:
Those changes are included in https://github.com/abartov/bybeconv/pull/406

PR #406, opened 2024-08-06
Collection toc

damisul:
Initial implementation of Collection-based TOC.

Now there is a new link 'New TOC' on authority page:

изображение

It will lead to a new page. It displays authority's TOC as a set of three nested ordered lists: for Original, Translated and Edited works. Every subcollection will add a new level of nesting to the ordered list:

изображение

I've also moved link to refresh uncollected works collection to new_toc page.

I've added caching of generated TOC page for 12 hours.

Also this PR includes validation check to prevent cyclic collection dependencies.

P.S. I've did not added I18n resource for New_toc as I expect this to be renamed soon (as well as controller action itself)

### DIFF
--- a/app/views/authors/_toc_node.html.haml
+++ b/app/views/authors/_toc_node.html.haml
@@ -1,0 +1,15 @@
+%li
+  - if toc_node.is_a?(Manifestation)
+    - if toc_node.published?
+      = link_to toc_node.title, manifestation_path(toc_node)
+    - else
+      = toc_node.title
+  - else
+    -# Collection node
+    - collection = toc_node.item
+    = link_to collection.title, collection_path(collection)
+    - children = toc_node.children_by_role(role, authority_id)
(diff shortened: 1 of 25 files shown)


app/views/authors/_toc_node.html.haml
```diff
@@ -0,0 +1,15 @@
+%li
+  - if toc_node.is_a?(Manifestation)
+    - if toc_node.published?
+      = link_to toc_node.title, manifestation_path(toc_node)
```
abartov, line 4: Here we need more nuance: if the node is an article in an issue of a periodical, or an item in an anthology by multiple authors, we want the name of the author(s) as well right here at the item level. We probably also want an indication (perhaps by icon) of the item's genre.

Precise UX for this screen is TBD, but I'm noting that this logic won't be enough.

abartov (approved):

PR #412, opened 2024-08-16
Collection CRUD and InvolvedAuthorities refactoring

damisul:
This PR contains several topics:
- modified involved_authorities table to have polymorphic relation item instead of three fields colleciton/work/expression
- added unique constraint to involved_authorities to prevent having same authority specified to same item with same role twice. Migration adds unuque index, and cleans db from possible duplicates
- updated UI for editing manifestation's metadata to allow adding involved_authorities without submitting whole form, and make this UI reusable. Also now it uses autocompletion to search for an authority instead of preloaded drop-down with full authors list.
- updated Collection new and edit form to use simple_form for UI and fixed existing bugs there. Added autocompletion field to select publication.
- Added ability to edit involved_authorities linked to Collection.


app/controllers/involved_authorities_controller.rb
```diff
@@ -2,17 +2,38 @@
 
 # Controller to work with InvolvedAuthority records
 class InvolvedAuthoritiesController < ApplicationController
-  before_action :require_editor
+  before_action except: %i(index) do |c|
```
abartov, line 5: can you explain this exception?

abartov (approved):

PR #442, opened 2024-11-26
Rspec fix

damisul:
Fixed error in specs.
Problem occured when there where two manifestations with same impressions_count.

In this case TextAPI uses additional sorting by id, but spec code ignored it. Added missing sorting by id

### DIFF
--- a/app/api/v1/texts_api.rb
+++ b/app/api/v1/texts_api.rb
@@ -47,7 +47,7 @@ class V1::TextsAPI < V1::ApplicationApi
       post do
         ids = params[:ids]
         records = ManifestationsIndex.find(ids)
-        # If ids contains one one value, Chewy returns single object instead of array, so explicitely convert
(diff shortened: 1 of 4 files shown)


app/api/v1/texts_api.rb
```diff
@@ -47,7 +47,7 @@ class V1::TextsAPI < V1::ApplicationApi
       post do
         ids = params[:ids]
         records = ManifestationsIndex.find(ids)
-        # If ids contains one one value, Chewy returns single object instead of array, so explicitely convert
+        # If ids contains only single value, Chewy returns single object instead of array, so explicitely convert
```
damisul, line 50: fixed typo

docker/Dockerfile
```diff
@@ -1,7 +1,8 @@
-FROM ruby:3.2.1-buster
+FROM ruby:3.3.5-bullseye
```
damisul, line 1: bumped version of docker image to match updated ruby version

abartov (approved):

PR #444, opened 2024-11-27
Added search by Collections to site wide search.

damisul:
Also:
- renamed ManifestationsSearch to SiteWideSearch
- added automatic reindexing of DictionaryEntry records

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -7,11 +7,8 @@ class SearchController < ApplicationController
   def results
     begin
       @searchterm = params[:search].nil? ? sanitize_term(params[:q]) : sanitize_term(params[:search])
-      unless params[:search].nil?
-        @search = ManifestationsSearch.new(@searchterm)
-      else
-        @search = ManifestationsSearch.new(query: @searchterm)
-      end
+      @search = params[:search].present? ? SiteWideSearch.new(@searchterm) : SiteWideSearch.new(query: @searchterm)
+
       @results = @search.search.page(params[:page])
       page = (params[:page] || 1).to_i
       @offset = (page - 1) * Kaminari.config.default_per_page
--- a/app/lib/site_wide_search.rb
+++ b/app/lib/site_wide_search.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require 'multi_index_search_request'
-class ManifestationsSearch
+# Class representing request to search whole website (texts, people, etc.) by string query
+class SiteWideSearch
   include ActiveData::Model
 
   attribute :query, type: String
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -148,6 +148,7 @@ group :development do
 end
 
 group :test, :development do
+  gem 'byebug'
```
damisul, line 151: I use this gem for debugging

app/controllers/search_controller.rb
```diff
@@ -7,11 +7,8 @@ def index
   def results
     begin
       @searchterm = params[:search].nil? ? sanitize_term(params[:q]) : sanitize_term(params[:search])
-      unless params[:search].nil?
-        @search = ManifestationsSearch.new(@searchterm)
-      else
-        @search = ManifestationsSearch.new(query: @searchterm)
-      end
+      @search = params[:search].present? ? SiteWideSearch.new(@searchterm) : SiteWideSearch.new(query: @searchterm)
```
damisul, line 10: I did not found any places where 'search' param is passed here. So perhaps we can remove this code branch?

app/lib/site_wide_search.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require 'multi_index_search_request'
-class ManifestationsSearch
+# Class representing request to search whole website (texts, people, etc.) by string query
+class SiteWideSearch
```
damisul, line 5: I've renamed class as its name was misleading

app/lib/site_wide_search.rb
```diff
@@ -101,8 +108,4 @@ def tags_filter
   def highlight
     index.highlight(max_analyzed_offset: 999_000, fields: { fulltext: {}, deftext: {} })
   end
-
```
damisul, line 104: This code was add probably to resolve items ordering issue, but after we implemented script-based ordering it become obsolete, so I've removed it

app/views/search/results.html.haml
```diff
@@ -27,6 +27,9 @@
             - unless result._data.nil? or result._data['highlight'].nil?
               %br
               != result._data['highlight']['fulltext'].join("<br />").gsub('<em>','<b>').gsub('</em>','</b>')
+          - elsif result.instance_of?(CollectionsIndex)
+            %b= Collection.model_name
+            %b= link_to([result.title, result.subtitle].compact_blank.join(' / '), collection_path(result.id))
```
damisul, line 32: Probably we may need to add there some tweaks for right-to-left text orientation. Also not sure if subtitle should be displayed at all. I simply did some basic view, we can update later.

Gemfile
abartov: So did I, but the latest and recommended debugging gem is 'debug'.

spec/controllers/search_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      create(:authority, name: 'Test Authority')
+      create(:manifestation, title: 'Test Manifestation')
+      create(:collection, title: 'Test Collection')
+    end
+  end
+
+  describe '#results' do
+    subject { get :results, params: { q: 'Test' } }
+
+    it { is_expected.to be_successful }
```
abartov, line 18: Please add testing of result count and order.

abartov (changes requested):
see inline comments.

app/controllers/search_controller.rb
abartov: Could be a remnant of some older code, yes.

abartov (approved):

PR #455, opened 2024-12-19
Sub-collection addition failed on collection/manage page

damisul:
It was impossible to add a new existing sub-collection (using autocomplete) on collection/manage page as we missed 'Collection' item_type value in drop-down.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -327,6 +327,7 @@ module ApplicationHelper
               .map { |k, _v| [textify_collection_type(k), k] } +
       [
         [t(:work), 'Manifestation'],
+        [Collection.model_name, 'Collection'],
         [t(:paratext), 'paratext'],
         [t(:placeholder_item), 'placeholder_item']
       ]


abartov:
Just FYI: I merged this by mistake. The behavior had in fact been correct: it is impossible to just add a Collection -- the method already includes all Collection types (volume, series, etc.) plus the leaf types. The user should pick a *specific* Collection type (e.g. 'volume') to add.  I have now removed the line you added.

damisul:
> Just FYI: I merged this by mistake. The behavior had in fact been correct: it is impossible to just add a Collection -- the method already includes all Collection types (volume, series, etc.) plus the leaf types. The user should pick a _specific_ Collection type (e.g. 'volume') to add. I have now removed the line you added.

Hi @abartov 
Maybe I've fixed it wrongly, but I still believe there is something to be fixed.

The problem I've tried to fix was not **creation of new Collection** as sub-collection, but adding some already existing Collection as a sub-collection. There is an autocomplete field to select collection, it allows to select existing works, but if we try to add it, an error occurs (it does not displays, but is visible in logs):


And if we'll check this line: https://github.com/abartov/bybeconv/blob/c508ec43fe45563012e8e20b3879ca6f4f01a5b4/app/views/shared/_manage_collection.html.haml#L105

We see that item_type == 'Collection' is also expected.

So I've asssumed that intended logic was:
- if we want to create new sub-collection and add it then we should select some collection_item type from drop-down (e.g. volume or periodical)
- but if we want to find and add existing collection, then item_type should be set to 'Collection' and we miss this value in a drop-down. So I've added it.

Without this entry we cannot add existing collection as sub-collection.

Sidethinking: I believe this UI for adding new Collection_item using single form is a bit contr-intuitive.
The problem with UI is that we need to select specific values in a drop-down for each possible case. I'd suggested to use several smaller forms instead:
- to add existing manifestation (using autocomplete)
- to add existing collection (using autocomplete)
- to create new placeholder
- to create new collection  (should only allows to select valid collection_type values from a drop-down without 'Collection' and 'Manifestation' entries)


So we need to at least fix this error then.

PR #538, opened 2025-06-10
Docker and Environment Settings

damisul:
This PR implements sevral things related to configuration and docker usage:

1. It moves docker-compose file used for development from docker/bybe_dev folder to root application folder. I've also removed dockerfile for application itself, as I did not used it and only used services like db/elastic, etc.
2. It updates application configuration to specify ElasticSearch and MySQL connection settings via two env variables: DATABASE_URL and ELASTICSEARCH_HOST. I've also added 
3. It removed Vagrantfile, as we use docker instead.

NOTE: before deploying app to production we need to define in host system two ENV variables: DATABASE_URL and ELASTICSEARCH_HOST. Values from `.env.development` and `.env` files can be used as the reference


README.docker.md
```diff
@@ -0,0 +1,141 @@
+This document describes our development docker configuration.
+
+> Instruction below are written for Debian linux. For other distros/OSes it can require some changes
+
+We assume that application itself will be run in host system natively, docker is used to only host services used by app:
```
abartov, line 5: In production, we use a separae AWS RDS service for the main production database, and not a Mysql instance on localhost.

abartov: @damisul ?

damisul:
Hey, @abartov !
Adapter should be specified as the first part of DATABASE_URL:

E.g. in our `.env.devlopment`
```
mysql2://root:password@127.0.0.1:13306/bybe_dev
```

<ADAPTER>://<USER>:<PASSWORD>@<DB_ADDRESS>:<DB_PORT>/<DB_NAME>

PR #539, opened 2025-06-15
Clean old base users

damisul:
- Created service to clean up stale base_users.
- Enchanced db:sessions:trim rake task to call this serice after each run

I've 'enchanced' `db:sessions:trim` rake task, to call BaseUsersCleanup after every call, so no need to schedule any new job.

BUT, there is is a problem with a first run. As we need to delete ~8 millions of records from BaseUsers table it takes a while. On my machine with MySQL running in docker it took ~30 minutes.  The problem is that Rails kills transaction due to lock timeout.
And it would be good to not have app running during first run.

To avoid this I propose following workaround:
1. Stop the app
2. Temporarily add
```
timeout: 1000
```
to production connection in `database.yml` (if it still fail with LockTimeout we'll need to increase it, but 1000 worked for me).
3. Run rake `db:sessions:cleanup` once
4. Remove timeout settings from `database.yml`
5. Start app again.

All following invocations of rake task should be quick, as there will not be so much records for deletion.

@abartov , FYI

### DIFF
--- a/app/services/clean_up_base_users.rb
+++ b/app/services/clean_up_base_users.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# This service deletes all base_user records for anonymous users where session record does not exists
+# We clean sessions table periodically by running standard rake task 'db:sessions:trim' this service should be
+# run after each sessions cleanup
+class CleanUpBaseUsers < ApplicationService
+  def call
+    stale_users = BaseUser.where(user: nil)
+                          .where('not exists (select 1 from sessions s where s.session_id = base_users.session_id)')
+
+    Bookmark.where(base_user: stale_users).delete_all
+
+    run_sql("delete from base_user_preferences where base_user_id in (#{stale_users.select(:id).to_sql})")
+    stale_users.delete_all
+
+    run_sql('optimize table base_users')
+  end
+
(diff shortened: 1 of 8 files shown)


app/services/clean_up_base_users.rb
```diff
@@ -0,0 +1,27 @@
...
+# We clean sessions table periodically by running standard rake task 'db:sessions:trim' this service should be
+# run after each sessions cleanup
+class CleanUpBaseUsers < ApplicationService
+  def call
+    stale_users = BaseUser.where(user: nil)
+                          .where('not exists (select 1 from sessions s where s.session_id = base_users.session_id)')
+
+    Bookmark.where(base_user: stale_users).delete_all
+
+    run_sql("delete from base_user_preferences where base_user_id in (#{stale_users.select(:id).to_sql})")
```
damisul, line 13: Probably after initial run we can rewrite job to 
```
stale_users.each(&:destroy!)
```
It will be much slower, but on small datasets should work fine

But for deletion of 8 millions of records it would be too slow.

abartov (approved):

abartov:
Now done in production. We're down to ~500K BaseUser records. Thanks!

PR #551, opened 2025-06-25
Replaced Impressionist gem with ahoy. Made Anthologies and Collections trackable

damisul:
This PR removes impressionist gem from project and adds Ahoy calls where it was used.

Key points of new logic:
- I've wiped all existing ahoy_events and ahoy_visits to start from the scratch
- Ahoy allows to specify arbitrary names for events, so we can split different actions, currently I've implemented `view` and `download` events (previously manifestation downloads was counted as view).
- I've kept old `impressions_count` column in Authorities and Manifestations table, but now it is simply increased by 1 on every track_view call. We can add similar `downloads_count` column if necessary in future. Also we can consider to rename `impressions_count` to `views_count`, but I've kept old name for now, as it is exposed via API.
- I've added Concern `Tracking` to include all logic used to track events from controllers. There are two methods:
  - `track_view`(object) - used to track view event (also increased impressions_count by 1)
  - `track_download`(object) - used to log download event
  - there is also `track_event`(event, object) method, but it should not be used directly (maybe in future when we'll add more events)
- all methods accepts object which will be associated with event (e.g. Authority or Manifestation), and we store it's type and id in ahoy properties field,
- for convenience I've added two 'virtual' column `item_id` and `item_type` to `ahoy_events` table, so now we can associate Ahoy::Events with referenced objects with `has_many` (see `TrackingEvents` concern)

Existing data from 2025 from impressions table are compacted and loaded into year_totals table for 2025 (there is a migration for it).

I've updated task used to compact data to:
- use data from ahoy_events instead of impressions
- update existing year_totals records if they are found (it adds to existing total value)
- also this task is renamed from `impressions:compact` to `ahoy:compact`
- also I've added 'event' column 

And as a final step I've added tracking to anthologies and collections.
It required adding impressions_count column to them, so we'll need to reset CollectionIndex (I've added impressions_count there as well).

### DIFF
--- a/app/controllers/concerns/tracking.rb
+++ b/app/controllers/concerns/tracking.rb
@@ -28,7 +28,8 @@ module Tracking
     return if spider?
 
     Ahoy::Event.transaction do
-      track_event('view', record)
+      props = object_properties_for_tracking(record)
+      track_event('view', props)
       record.increment!(:impressions_count) # we simply increase impressions_count here
 
       # increment! method does not trigger Chewy index update so we do it explicitly here
@@ -45,8 +46,10 @@ module Tracking
 
   # Download event should be triggered when we download file associated with object
   # @param object being downloaded (Manifestation, Collection, etc)
-  def track_download(record)
-    track_event('download', record)
+  def track_download(record, format)
+    props = object_properties_for_tracking(record)
+    props[:format] = format
+    track_event('download', props)
   end
 
   # Page view event is used to track generic page views
@@ -57,31 +60,25 @@ module Tracking
 
   # Generic method used to track random events
   # @param event Event name to be tracked (Ensure it is included in Ahoy::Events::ALLOWED_NAMES)
-  # @param record can be either ActiveRecord instance or Hash, specifies additional params to be stored in
-  #   event properties. If record is an ActiveRecord it will add `id` and `type` attributes to properties, if record
-  #   is a Hash it will simply use them.
-  # NOTE: this method always adds to `properties` action and controller names, so no need to add them exlicitely
-  def track_event(event, record = nil)
+  # @param properties Hash object containing additional properties to associate with event
+  # NOTE: this method always adds to `properties` action and controller names, so no need to add them explicitely
+  def track_event(event, properties = {})
     raise "Unknown event: #{event}" unless Ahoy::Event::ALLOWED_NAMES.include?(event)
 
     return if spider?
 
-    props = if record.nil?
-              {}
-            elsif record.is_a?(ActiveRecord::Base)
-              {
-                type: record.class.name,
-                id: record.id
-              }
-            elsif record.is_a?(Hash)
-              record
-            else
-              raise 'record must be a Hash or ActiveRecord object'
-            end
+    properties[:controller] = controller_name
+    properties[:action] = action_name
 
-    props[:controller] = controller_name
-    props[:action] = action_name
+    ahoy.track(event, properties)
+  end
 
-    ahoy.track(event, props)
+  def object_properties_for_tracking(object)
+    raise 'Object must be an ActiveRecord object' unless object.is_a?(ActiveRecord::Base)
+
+    {
+      type: object.class.name,
+      id: object.id
+    }
   end
 end
--- a/app/controllers/manifestation_controller.rb
+++ b/app/controllers/manifestation_controller.rb
@@ -301,9 +301,8 @@ class ManifestationController < ApplicationController
     # Without this we had situation when Downloadable object was created but attachmnt creation failed
     Downloadable.transaction do
       m = Manifestation.find(params[:id])
-      track_download(m)
-
       dl = GetFreshManifestationDownloadable.call(m, format)
+      track_download(m, format)
       redirect_to rails_blob_url(dl.stored_file, disposition: :attachment)
     end
   end
--- a/app/controllers/anthologies_controller.rb
+++ b/app/controllers/anthologies_controller.rb
@@ -82,7 +82,7 @@ class AnthologiesController < ApplicationController
         end
         dl = MakeFreshDownloadable.call(params[:format], filename, html, @anthology, austr)
       end
-      track_download(@anthology)
+      track_download(@anthology, format)
       redirect_to rails_blob_url(dl.stored_file, disposition: :attachment)
     else
       redirect_to '/', error: t(:no_permission)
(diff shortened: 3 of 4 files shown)


app/controllers/concerns/tracking.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      # increment! method does not triggers Chewy index update so we do it explicitely here
+      # for performance purpose we specify only single field to be updated
+      index_class = case record.class.name
+                    when 'Collection' then CollectionsIndex
+                    when 'Manifestation' then ManifestationsIndex
+                    when 'Authority' then AuthoritiesIndex
+                    end
+
+      index_class&.send(:import, record, import_fields: [:impressions_count])
```
abartov, line 40: I am concerned about performance here: would this not mean that a popular book-length text would cause very frequent re-indexing of the whole book fulltext _on each view?_
The impressions_count field doesn't need to be accurate in real-time, and can definitely be updated, say, once every 24 hours. So I suggest adding a check here comparing current time to the record's DB updated_at value, and only updating the ES index if it's been more than 24h.

app/controllers/manifestation_controller.rb
```diff
@@ -840,45 +836,31 @@ def refuse_unreasonable_page
...
+    @e = @m.expression
+    @w = @e.work
+    @author = @w.authors[0] # TODO: handle multiple authors
+    track_view(@m)
+    track_view(@author)
+
+    if @author.nil?
+      @author = Authority.new(name: '?')
+    end
+    @translators = @m.translators
```
abartov, line 855: I would also like to track a view for every involved translator. It's an oversight that we weren't.
(but no need for editors, illustrators, and other roles.)

app/controllers/concerns/tracking.rb
damisul: Hey @abartov , well, I believe this version should be much faster compared to existing approach.
Currently we use `save` method to increase `impressions_count` see: https://github.com/abartov/bybeconv/blob/0ca59ef857615924ed5f69da7369bd00b5685e09/app/models/application_record.rb#L11

New approach implemented in this PR has following performance preferences:
- `increment!` does not run validation on model, but `save` validates whole model
- `increment!` simply increase value of column by 1, but old implementation fetched counts from `year_totals` and added to then count of not yet compacted impressions. Two additional SQL calls on pretty large tables.
- `save` triggered whole object reindex (all columns), while new implementation explicitely says to only update `impressions_count` field in index

>  So I suggest adding a check here comparing current time to the record's DB updated_at value, and only updating the ES index if it's been more than 24h.

`increment!` does not updates `updated_at` (and if you check old implementation it also intentionally did not updated it, see https://github.com/abartov/bybeconv/blob/0ca59ef857615924ed5f69da7369bd00b5685e09/app/controllers/authors_controller.rb#L592. So we cannot rely on update_at.

As we already updated elasticsearch on every view, I believe it is OK to keep it updated. And new implementation should be faster.

app/controllers/manifestation_controller.rb
```diff
@@ -840,45 +836,31 @@ def refuse_unreasonable_page
...
+      flash[:notice] = t(:work_not_available)
+      redirect_to '/'
+      return
     end
+
+    @e = @m.expression
+    @w = @e.work
+    @author = @w.authors[0] # TODO: handle multiple authors
+    track_view(@m)
+    track_view(@author)
```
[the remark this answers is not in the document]
damisul: Ok, done

app/chewy/collections_index.rb
```diff
@@ -12,4 +12,5 @@ class CollectionsIndex < Chewy::Index
   field :involved_authorities_string, value: lambda { |c|
     c.involved_authorities.preload(:authority).map { |ia| ia.authority.name }.join(', ')
   }
+  field :impressions_count, type: :integer
```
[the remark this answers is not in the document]
abartov: Good! We will indeed be implementing a collections browsing screen, including filtering and sorting, very similar to /works

abartov (approved):

abartov (changes requested):
Please implement homepage and search events before we'll merge this.

app/services/clean_up_ahoy_visits.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# This service deletes all ahoy_visits created before given date and having no events logged
+# Typically this service should be called after ahoy_events cleanup
+class CleanUpAhoyVisits < ApplicationService
```
damisul, line 5: Extracted logic to clean up old Visits into a separate service.

app/controllers/welcome_controller.rb
```diff
@@ -1,7 +1,5 @@
 include BybeUtils
 class WelcomeController < ApplicationController
-#  impressionist # log actions for pageview stats
-
```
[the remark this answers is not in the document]
damisul: Done

abartov (approved):

abartov:
On running db:migrate on my dev DB after pulling from master, I got this:

```
== 20250624185016 DropImpressions: migrating ==================================
-- execute("insert into year_totals (item_type, item_id, year, total, event, created_at, updated_at)\nselect impressionable_type, impressionable_id, year(updated_at) as year, count(*), 'view', current_timestamp(), current_timestamp() \nfrom impressions \ngroup by impressionable_id, impressionable_type, year(updated_at)\n")
rails aborted!
StandardError: An error has occurred, all later migrations canceled: (StandardError)

Mysql2::Error: Duplicate entry 'Authority-43-2024' for key 'year_totals.index_year_total_by_year'
/home/asaf/dev/bybeconv/db/migrate/20250624185016_drop_impressions.rb:5:in `change'

Caused by:
ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry 'Authority-43-2024' for key 'year_totals.index_year_total_by_year' (ActiveRecord::RecordNotUnique)
/home/asaf/dev/bybeconv/db/migrate/20250624185016_drop_impressions.rb:5:in `change'

Caused by:
Mysql2::Error: Duplicate entry 'Authority-43-2024' for key 'year_totals.index_year_total_by_year' (Mysql2::Error)
/home/asaf/dev/bybeconv/db/migrate/20250624185016_drop_impressions.rb:5:in `change'
```

abartov:
Now successfully deployed in production.

PR #554, opened 2025-07-01
Rails 7

damisul:
Ported application to rails 7

Some files had to be renamed to match class names due to new classloader.
Another big change is that in newer version of Rails belongs_to is mandatory by default (actually it was like this even in 6.1 but we had this option disabled). So I had to update number of `belongs_to` declaration and add there `optional: true`.


app/models/collection.rb
```diff
@@ -451,20 +451,20 @@ def invalidate_cached_credits!
...
-      ci.seqno = before_seqno + 1
-      collection_items[pos - 1..].each do |coli|
-        coli.seqno += 1
-        coli.save!
-      end
-      ci.save!
-      ret = ci.id
-    end
-    return ret
+    new_seqno = if pos <= 1
```
damisul, line 454: There were a bug in this method so I've rewritten it

app/models/html_file.rb
```diff
@@ -325,16 +325,16 @@ def validate(record)
 end
 
 class HtmlFile < ApplicationRecord
-  include Ensure_docx_content_type # fixing docx content-type detection problem, per https://github.com/thoughtbot/paperclip/issues/1713
+  include EnsureDocxContentType # fixing docx content-type detection problem, per https://github.com/thoughtbot/paperclip/issues/1713
```
damisul, line 328: New classloader requires filename to match class name (class name should be in CamelCase, and filename in snake_case)

app/api/v1/api.rb
```diff
@@ -1,14 +1,19 @@
...
+    version :v1, using: :path
+    content_type :json, 'application/json'
+    default_format :json
 
-  prefix :api
+    prefix :api
 
-  mount V1::TextsAPI => '/'
-  mount V1::AuthoritiesAPI => '/'
+    mount TextsApi => '/'
```
damisul, line 14: Had to rename API classes to match new naming convention: API -> Api

app/models/aboutness.rb
```diff
@@ -1,5 +1,5 @@
 class Aboutness < ApplicationRecord
   belongs_to :work # work that is ABOUT something
   belongs_to :user
-  belongs_to :aboutable, polymorphic: true # something that the work is ABOUT
+  belongs_to :aboutable, polymorphic: true, optional: true # something that the work is ABOUT
```
damisul, line 4: Here and below. Now by default all belong_to are mandatory, so we need to add `optional: true` where this relation is optional

app/models/tagging.rb
```diff
@@ -2,16 +2,13 @@
...
 # Maps tags to other system object (Authorities, Manifestations, etc)
 class Tagging < ApplicationRecord
-
-  belongs_to :tag, foreign_key: 'tag_id', counter_cache: true
+  belongs_to :tag, counter_cache: true
   belongs_to :taggable, polymorphic: true # taggable things include Manifestations, People, Anthologies, ...
 
   belongs_to :suggester, foreign_key: 'suggested_by', class_name: 'User'
-  belongs_to :approver, foreign_key: 'approved_by', class_name: 'User'
-  validates :suggester, presence: true
```
damisul, line 11: Here and below: this validation become redundant as belongs_to is mandatory be default

config/environments/development.rb
```diff
@@ -1,18 +1,25 @@
+require "active_support/core_ext/integer/time"
```
damisul, line 1: I've regenerated all env files using Rails 7 defaults and then backported all changes from old files

This will simplify future upgrades

config/locales/active_record.he.yml
```diff
@@ -19,6 +19,8 @@ he:
       featured_content: תוכן מערכת
       person: אדם
     errors:
+      messages:
+        required: required
```
damisul, line 23: need translation (this message is shown when mandatory field is missing in ActiveRecord object)

spec/models/collection_spec.rb
```diff
@@ -177,28 +179,29 @@
...
       end
     end
 
     context 'when insert at specified position' do
       let(:index) { 2 }
 
       it 'inserts successfully' do
-        expect(collection.collection_items.count).to eq 4
-        expect(collection.collection_items.map(&:item)).to eq(
-          [first_item.item, manifestation, third_item.item, second_item.item]
```
damisul, line 200: There were a bug, but it passed unnoticed as all collection_items created by factory had nulls items. Probably it was one of the problems with items reordering mentioned here: https://github.com/abartov/bybeconv/issues/529

app/models/aboutness.rb
damisul: Het @abartov 
For most fields where I was not sure I've set field as `optional` only if we have records in database with this field being `null`.

In this particular case it is clearly valid, as we can have `Aboutness` referencing `wikidata_qid` instead of `Manifestation` or `Authority`.

abartov: oh right! I forgot about that. :)

abartov (approved):

PR #565, opened 2025-07-11
Proofs for Collections

damisul:
- dropped deprecated Proofs.html_file_id column
- made Proofs polymorphic (for now still only used for Manifestations)
- implemented proofs reporting from the collection show page.

Now we need to wrap text to be proofread with the div with class 'proofable' and data attributes defining id and type of the item to be proofread. I've added it to both collections/read and manifestation/read pages, so this approach should be universal and can be used for other entities as well in future. Also this approach will check if user actually selected text from the work (previously he could select any random text on the page).


app/controllers/proofs_controller.rb
```diff
@@ -42,79 +50,83 @@ def index
...
-      @proofs = @proofs.joins(:manifestation).where('manifestations.title LIKE ?', "%#{@search_query}%")
+      @proofs = @proofs.where(
+        <<~SQL.squish,
+          exists (
+            select 1 from
+              manifestations m
+            where
+              m.id = proofs.item_id
+              and proofs.item_type = 'Manifestation'
+              and upper(m.title) like ?
```
damisul, line 67: made filtering case-insensitive

app/controllers/proofs_controller.rb
```diff
@@ -42,79 +50,83 @@ def index
     @search_query = params[:search]
 
     @proofs = if @status.nil?
-                Proof.where.not(status: :spam).order(:manifestation_id)
+                Proof.where.not(status: :spam)
               else
-                Proof.where(status: @status).order('updated_at DESC')
+                Proof.where(status: @status)
```
abartov, line 55: here, too, ordering was useful -- since this is usually used to review *resolved* proofs (to undo a mistaken resolution, or to double-chech something), in the vast majority of cases, the desired proofs are some of the latest-resolved ones.

db/migrate/20250710064800_make_proofs_polymorphic.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    change_column_null :proofs, :item_id, false
+    change_column_null :proofs, :item_type, false
+
+    remove_index :proofs, [:status, :manifestation_id]
+    remove_column :proofs, :manifestation_id
+
+    add_index :proofs, [:item_id, :item_type, :status]
+    add_index :proofs, [:item_type]
+  end
```
abartov, line 22: what about an index by updated_at?

abartov (changes requested):
Looks good overall. Left some comments.

db/migrate/20250710064800_make_proofs_polymorphic.rb
damisul: Well... Index by updated_at itself is will not be useful I believe.
We only sort by updated_at when filter by status is applied. And we do have index by status already.
We can add index on both colums (status and updated_at) but I believe it is premature optimization

abartov (approved):

PR #566, opened 2025-07-12
Added Collections#fix_ordering action

damisul:
Added fix_ordering button to collections manage page to fix situations when we have non-unique seqno in collection items.
Such situations make it impossible to correctly reorder items, so we need a way to fix it.

It adds new button to manage page. After clicking on it system will ensure all items in colleciton (and nested subcollections) has unique sequential `seqno` values (starting from 1). So it will also removes all gaps in `seqno` values.

if some collision was found it will display, corresponding message, otherwise it will say that no problems were found (like on screenshot):
screenshot

NOTE: some new translations to locale file are required.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -6,7 +6,7 @@ class CollectionsController < ApplicationController
   include Tracking
 
   before_action :require_editor, except: %i(show download print)
-  before_action :set_collection, only: %i(show edit update destroy drag_item)
+  before_action :set_collection, only: %i(show edit update destroy drag_item fix_ordering)
 
   # GET /collections or /collections.json
   def index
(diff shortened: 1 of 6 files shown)


abartov:
Wait, why do we want this button? How would users know -- given they don't see seqnos at all -- that it's time to use the button?

If non-unique seqnos have crept in due to previous bugs, let's run a single migration fixing them in production. Do we expect it to happen again?

damisul:
well, as I've wrote in linked issue I feel there could be some not yet found bugs which may break seqno uniqueness. That's why I've decided to add button. 

I can create rake task instead, but it will require you to run it manually. Not sure what is more convenient.

We can for example hide this button from regular editors and make it on;y available for admins.

abartov:
Ah, sorry, I missed the comments in the issue. Okay, understood, but no, I don't think a manual button is the way to go. Please re-implement as a rake task. I think it would be a good idea to have it run nightly in production, and please make it send an email to editor@benyehuda.org with the fix count **if** it is nonzero. That way, we can keep track of how often (or not) duplicate seqnos are created, and either conclude we are done, or keep hunting for another bug.

damisul:
Hey, @abartov ! I've rewritten PR to use Rake task approach. Please re-review

abartov (approved):

PR #570, opened 2025-07-17
Rails 7.1

damisul:
Bumped version of Rails to 7.1

I've re-generated environment files once again and backported changes from our configs there.

Also there are three migrations to update active-storage related tables, they were generated by rails app:update script.


app/controllers/authors_controller.rb
```diff
@@ -12,7 +12,7 @@ class AuthorsController < ApplicationController
   end
 
   before_action :set_author,
-                only: %i(show edit update destroy toc edit_toc prep_for_print print all_links delete_photo
+                only: %i(show edit update destroy toc edit_toc print all_links delete_photo
```
damisul, line 15: removed not existing `prep_for_print` action

abartov (approved):

PR #572, opened 2025-07-20
upgraded grape from 1.6.0 to 2.4.0

damisul:
removed grape_extra_validators dependency and replaced it with new standard validators, introduced in grape 2.x

I've did a quick test and it seems to work. The only difference is that wording in a case of validation error is changed (translation required)

### DIFF
--- a/app/api/v1/texts_api.rb
+++ b/app/api/v1/texts_api.rb
@@ -41,7 +41,7 @@ module V1
         params do
           requires :ids,
                    type: [Integer],
-                   maximum_length: 25,
+                   length: { max: 25 },
                    allow_blank: false,
                    desc: 'array of text IDs to fetch',
                    documentation: { param_type: 'body' }
--- a/app/api/v1/validations/auth_key.rb
+++ b/app/api/v1/validations/auth_key.rb
@@ -3,7 +3,7 @@
 module V1
   module Validations
     # Auth key check
-    class AuthKey < Grape::Validations::Base
+    class AuthKey < Grape::Validations::Validators::Base
       class AuthFailed < StandardError; end
 
       def validate_param!(attr_name, params)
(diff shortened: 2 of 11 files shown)


app/api/v1/texts_api.rb
```diff
@@ -41,7 +41,7 @@ class TextsApi < V1::ApplicationApi
         params do
           requires :ids,
                    type: [Integer],
-                   maximum_length: 25,
+                   length: { max: 25 },
```
damisul, line 44: standard validator introduced in grape 2.0 is used now

app/api/v1/validations/auth_key.rb
```diff
@@ -3,7 +3,7 @@
 module V1
   module Validations
     # Auth key check
-    class AuthKey < Grape::Validations::Base
+    class AuthKey < Grape::Validations::Validators::Base
```
damisul, line 6: base class for custom validators has been changed

config/locales/grape.he.yml
```diff
@@ -0,0 +1,5 @@
+he:
+  grape:
+    errors:
+      messages:
+        length_max: 'is expected to have length less than or equal to %{max}'
```
damisul, line 5: Grape uses i18n now, but by default only supports english. I've added file for Hebrew (only one message for now)

spec/api/v1/authorities_api_spec.rb
```diff
@@ -9,17 +9,15 @@
     subject(:call) { get path }
 
     let(:detail) { 'metadata' }
-    let(:authority_id) { -1 }
```
damisul, line 12: grape does not accepts negative ids in URL anymore, so I've had to update specs to use positive not-existing ids

PR #579, opened 2025-07-31
reorganized i18n of genre values

damisul:
While working on Lexicon branch I've faced one problem: we have `lexicon` genre and new lexicon-related entities.
It caused conflicts in i18n file, as genre values was declared as the top-level resources.

I've moved all genres i18n strings to be under new `genre_values` node and replaced all direct usage of i18n with `textify_genre` helper.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -360,7 +360,7 @@ class ApplicationController < ActionController::Base
     author.each do |genre|
       next unless genre[1].class == Array # skip the :latest key
 
-      worksbuf = "<strong>#{I18n.t(genre[0])}:</strong> "
+      worksbuf = "<strong>#{textify_genre(genre[0])}:</strong> "
       first = true
       genre[1].each do |m|
         title = m.expression.title
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -67,7 +67,7 @@ module ApplicationHelper
   def textify_genre(genre)
     return I18n.t(:unknown) if genre.nil? or genre.empty?
 
-    return I18n.t(genre)
+    return I18n.t("genre_values.#{genre}")
   end
 
   def textify_intellectual_property(value)
--- a/app/views/ingestibles/_form.html.haml
+++ b/app/views/ingestibles/_form.html.haml
@@ -73,7 +73,7 @@
           = f.hidden_field :status, value: @ingestible.status
         %h2= t('ingestible.defaults')
         = f.label t(:genre)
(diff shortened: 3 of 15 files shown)


app/controllers/application_controller.rb
```diff
@@ -360,7 +360,7 @@ def textify_new_pubs(author)
     author.each do |genre|
       next unless genre[1].class == Array # skip the :latest key
 
-      worksbuf = "<strong>#{I18n.t(genre[0])}:</strong> "
+      worksbuf = "<strong>#{textify_genre(genre[0])}:</strong> "
```
damisul, line 363: Here and below: we should not use `I18n.t(genre)`, we should use `textify_genre` helper instead.

app/helpers/application_helper.rb
```diff
@@ -67,7 +67,7 @@ def options_for_shelves
   def textify_genre(genre)
     return I18n.t(:unknown) if genre.nil? or genre.empty?
 
-    return I18n.t(genre)
+    return I18n.t("genre_values.#{genre}")
```
damisul, line 70: This method should be single point of thruth for i18n of genre values

app/views/ingestibles/_form.html.haml
```diff
@@ -73,7 +73,7 @@
           = f.hidden_field :status, value: @ingestible.status
         %h2= t('ingestible.defaults')
         = f.label t(:genre)
-        - genres = [[t(:mixed_genres), 'mixed']]+get_genres.map {|genre| [I18n.t(genre), genre]}
+        - genres = (['mixed'] + Work::GENRES).map { |genre| [textify_genre(genre), genre] }
```
damisul, line 76: I've also added `mixed` value to `genre_values` i18n node.

Sidenote: I've checked content of `Ingestibles` table and we have number of bad values there (when genre is stored in Hebrew). Not sure if we should fix this.

config/locales/he.yml
```diff
@@ -1587,7 +1577,17 @@ he:
...
+    poetry: שירה
+    prose: פרוזה
+    article: מאמרים ומסות
+    drama: מחזות
+    memoir: זכרונות ויומנים
+    letters: מכתבים
+    reference: עיון
+    lexicon: מילונים ולקסיקונים
+    fables: משלים
+    mixed: סוגות שונות
```
damisul, line 1590: this line is only used for Ingestion model

app/views/ingestibles/_form.html.haml
abartov: thank you for noticing! Must have been an early bug I had fixed without fixing the data. Now manually fixed in the production DB. You can fix (or just randomly assign) in yours.

PR #580, opened 2025-08-01
Rails 7.2

damisul:
Updated app to Rails 7.2

### Enums
Biggest change there is that in Rails 7.2 they deprecated old keyword-like syntax for declaring enums, so instead of

```ruby
enum status: { new: 0, in_work: 1, closed: 2 }, _prefix: true
```

we should use new syntax:
```ruby
enum :status, { new: 0, in_work: 1, closed: 2 }, prefix: true
```

Old syntax to be removed in Rails 8, so I've updated all our enums to use new syntax. Also I've updated all enum declarations 
where values was specified as Array to use Hash.

### Configs
Once again I've regenrated environment files to apply new defaults and backported there our settings.
It also regenerated some error pages as well and added native rubocop integration

### Rubocop
This version of Rails introduces built-in rubocop support. Also I had to update our rubocop version to support newer Rails versions.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -9,7 +9,8 @@ Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
   # While tests run files are not watched, reloading is not necessary.
-  config.enable_reloading = false
+  # But we enable reloading to support spec debugging from IDE
+  config.enable_reloading = true
 
   # Eager loading loads your entire application. When running a single test locally,
   # this is usually not necessary, and can slow down your test suite. However, it's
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -132,12 +132,12 @@ group :development do
...
-  gem 'pronto', '~> 0.11.2'
-  gem 'pronto-haml', '~> 0.11.1', require: false
-  gem 'pronto-rubocop', '~> 0.11.5', require: false
-  gem 'rubocop', '~> 1.61.0', require: false
-  gem 'rubocop-factory_bot', '~> 2.25.1', require: false
-  gem 'rubocop-rails', '~> 2.24.1', require: false
+  gem 'pronto'
+  gem 'pronto-haml', require: false
+  gem 'pronto-rubocop', require: false
+  gem 'rubocop', '1.79.1', require: false
```
damisul, line 138: bumped rubocop

PR #581, opened 2025-08-06
Authorities and Manifestation autocomplete tweaks

damisul:
- Modified all places where we use `autocomplete_authority_name` endpoint to use `autocomplete_authority_name_and_aliases`, and removed first endpoint.
- Updated autocompletion on manifestaion browse filter to also autocomplete by name and aliases, but also added filter by `authority.status` = `published` (I believe on browse form we should not show not yet published authors)
- Updated `autocomplete_manifestation_title` endpoint to also search by `alternate_titles` field

Also I've added a workaround for situation when we try to edit `FeaturedContent` record and it has `user` = null (we have few such records, probably a legacy ones). Such situation started to cause failures after upgrade to Rails 7+ (as `belongs_to` become mandatory). To fix that I set `user` to `current user` if we try to update `FeaturedContent` with `user` = null.

### DIFF
--- a/app/controllers/admin/featured_contents_controller.rb
+++ b/app/controllers/admin/featured_contents_controller.rb
@@ -25,16 +25,15 @@ module Admin
       end
     end
 
-    def show
-      return unless @fc.nil?
-
-      flash[:error] = I18n.t(:no_such_item)
-      redirect_to url_for(action: :index)
-    end
+    def show; end
 
     def edit; end
 
     def update
+      # We have few legacy records in DB where `user` is null, and it caused validation failures during update
+      # To avoid this, we set user to current_user if it is empty
+      @fc.user = current_user if @fc.user.nil?
+
       if @fc.update(fc_params)
         flash.notice = I18n.t(:updated_successfully)
         redirect_to admin_featured_content_path(@fc)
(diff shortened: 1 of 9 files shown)


app/controllers/admin/featured_contents_controller.rb
```diff
@@ -25,16 +25,15 @@ def create
       end
     end
 
-    def show
-      return unless @fc.nil?
```
damisul, line 29: This check is useless as we'll get `RecordNotFound` error if wrong `id` is provided. So I've removed it.

app/controllers/admin/featured_contents_controller.rb
```diff
@@ -25,16 +25,15 @@ def create
...
-      redirect_to url_for(action: :index)
-    end
+    def show; end
 
     def edit; end
 
     def update
+      # We have few legacy records in DB where `user` is null, and it caused validation failures during update
+      # To avoid this, we set user to current_user if it is empty
+      @fc.user = current_user if @fc.user.nil?
```
damisul, line 35: Fixing legacy records with `user` = nil

abartov (approved):

PR #589, opened 2025-08-15
Use ElasticSearch for autocomplete

damisul:
Added two lightweight ElasticSearch indices to perform autocomplete lookups on authorities names and manifestation titles (including aliases).

Updated our autocomplete logic to use it instead of SQL queries

NOTE: we'll need to do `rake chewy:reset` to build new indices


app/controllers/admin_controller.rb
```diff
@@ -737,7 +752,7 @@ def tag_review
     else
       stags = ListItem.where(listkey: 'tag_similarity', item: @tag).pluck(:extra).map do |x|
                 x.split(':')
-              end.sort_by { |score, tag| score }.reverse
+              end.sort_by { |score, _tag| score }.reverse
```
damisul, line 755: rubocop autocorrection

abartov (approved):

PR #613, opened 2025-09-25
Removed root collection type

damisul:
Removed all leftovers of `root` collection type and dropped `Authorites.root_collection_id` column.
We replaced it with autogenerated TOC tree.

P.S. I've checked my db copy beforehand and there were no existing records with 'root' type, but I would doublechecked on production

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_07_10_064800) do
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -727,7 +725,7 @@
     t.index ["task_id"], name: "index_publications_on_task_id"
   end
 
-  create_table "reading_lists", charset: "utf8mb4", collation: "utf8mb4_bin", force: :cascade do |t|
+  create_table "reading_lists", charset: "latin1", force: :cascade do |t|
```
abartov, line 728: absolutely no table should be 'latin1' in our DB.

PR #637, opened 2025-10-15
CI pipeline fixes

damisul:
Regarding setting up CI pipeline.

@abartov , I see you've added new workflow 'rubyonrails' and it tries to run three tools:

- brakeman
- bundler-audit
- rubocop

As far as I see first two are some kind of security checks (first one is a static analyzer, second one simply checks if bundled gems have some security updates.

Rubocop can be excluded as we use it in pronto workflow.


.github/workflows/rubyonrails.yml
```diff
@@ -11,61 +11,21 @@ on:
...
-          sudo apt-get update -yq && sudo apt-get install -yq --no-install-recommends mysql-client libmysqlclient-dev wkhtmltopdf pandoc yaz libyaz-dev  libmagickwand-dev libpcap-dev memcached cmake
-      - name: Install Ruby and gems
-        uses: ruby/setup-ruby@78c01b705fd9d5ad960d432d3a0cfa341d50e410 # v1.179.1
+          sudo apt-get update
+          sudo apt-get install -y mysql-client libmysqlclient-dev wkhtmltopdf pandoc yaz libyaz-dev libmagickwand-dev libpcap-dev cmake
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-      - name: Generate binstubs
```
damisul: removed this step, as I've added brakeman and bundle-audit to Gemfile as test/dev dependencies and binstubs for them included in bin folder.

.ruby-version
```diff
@@ -1 +1 @@
-3.3.2
```
damisul: returned back to 3.3.8

damisul:
@abartov , currently it fails on bundle-audit check, as it found update for an `uri` gem.
I've left this unresolved, so you can evaluate output of bundle-audit.

To fix this we simply need to update version of `uri` gem  and commit updated Gemfile.lock

PR #655, opened 2025-10-19
Fix sortable receive: use source previndex to avoid order corruption

abartov:
Fixes intermittent collection item order corruption when moving items between collections. The receive handler now reads the previous index from the source sortable (ui.sender) and uses parseInt(...,10). Adds unit tests (Jest) for the helper that computes old_pos/new_pos.


abartov:
Made a mess by starting it from the wrong branch. Cherry-picked the fix, and closing this PR.

PR #710, opened 2025-11-10
Resolved performance issues on welcome page while getting popular authors and works

damisul:
## Problems
- for popular manifestations it fetched all view events for last months and looped over them to get most popular works
- for authorities it was the same, but result was ignored and old logic was used.

## Changes

- updated both places to use SQL with group by to get 10 most viewed records
- refactored popular_authors page to use Rails Caching instead of memoizing in class variable
- added new index to ahoy_events table to further speedup process

### DIFF
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -340,32 +340,17 @@ class Authority < ApplicationRecord
                                            taggable_id: mm }).group('tags.id').order('count_all DESC').count
   end
 
-  # class variable
-  # rubocop:disable Style/ClassVars
-  @@popular_authors = nil
-
-  def self.recalc_popular
-    evs = Ahoy::Event.where(name: 'view').where("JSON_EXTRACT(properties, '$.type') = 'Authority'").where(
-      'time > ?', 1.month.ago
-    )
-    pop = {}
-    evs.each do |x|
-      au_id = x.properties['id']
-      pop[au_id] = 0 unless pop[au_id].present?
-      pop[au_id] += 1
-    end
-    sorted_pop_keys = pop.keys.sort_by { |k| pop[k] }.reverse
-    Authority.find(sorted_pop_keys[0..9])
-
-    @@popular_authors = has_toc.order(impressions_count: :desc).limit(10).all.to_a
-  end
-  # rubocop:enable Style/ClassVars
-
   def self.popular_authors
-    if @@popular_authors.nil?
-      recalc_popular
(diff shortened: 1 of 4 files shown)


app/models/authority.rb
```diff
@@ -340,32 +340,17 @@ def popular_tags_used_on_works_with_count
...
-    pop = {}
-    evs.each do |x|
-      au_id = x.properties['id']
-      pop[au_id] = 0 unless pop[au_id].present?
-      pop[au_id] += 1
-    end
-    sorted_pop_keys = pop.keys.sort_by { |k| pop[k] }.reverse
-    Authority.find(sorted_pop_keys[0..9])
-
-    @@popular_authors = has_toc.order(impressions_count: :desc).limit(10).all.to_a
```
damisul: And here we simply ignored result of previous loop and fetched authors based on impressions_count (i.e. using alltime stats)

app/models/authority.rb
```diff
@@ -340,32 +340,17 @@ def popular_tags_used_on_works_with_count
...
-  end
-  # rubocop:enable Style/ClassVars
-
   def self.popular_authors
-    if @@popular_authors.nil?
-      recalc_popular
+    Rails.cache.fetch('m_popular_authors', expires_in: 24.hours) do
```
damisul: Used Rails.cache

app/models/manifestation.rb
```diff
@@ -415,18 +415,15 @@ def self.cached_pd_count
...
-      )
-      pop = {}
-      evs.each do |x|
-        mid = x.properties['id']
-        pop[mid] = 0 unless pop[mid].present?
-        pop[mid] += 1
-      end
-      sorted_pop_keys = pop.keys.sort_by { |k| pop[k] }.reverse
-      Manifestation.find(sorted_pop_keys[0..9])
+    Rails.cache.fetch('m_popular_works', expires_in: 24.hours) do
```
damisul: Used same approach as for Authorites. I've also changed cache period to 24 hours instead of 48.

db/migrate/20251110022619_add_index_to_ahoy_events.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIndexToAhoyEvents < ActiveRecord::Migration[8.0]
+  def change
+    add_index :ahoy_events, [:time, :name, :item_type, :item_id]
```
damisul: Added new index to further speedup queries.

PR #712, opened 2025-11-10
Manifestation read speedup #1

damisul:
This is a first set of changes to improve performance of Manifestation#read action:

- removed code to fetch random work (it was not used)
- reorganized code used to fetch info about recommendations (to use fewer SQL queries)

### DIFF
--- a/app/controllers/manifestation_controller.rb
+++ b/app/controllers/manifestation_controller.rb
@@ -252,13 +252,11 @@ class ManifestationController < ApplicationController
   end
 
   def read
-    @m = Manifestation.joins(:expression).includes(:expression).find(params[:id].to_i)
-    if @m.nil?
-      head :not_found
-    elsif @m.expression.work.genre == 'lexicon' && DictionaryEntry.where(manifestation_id: @m.id).count > 0
+    @m = Manifestation.find(params[:id].to_i)
+    if @m.expression.work.genre == 'lexicon' && DictionaryEntry.exists?(manifestation: @m)
       redirect_to action: 'dict', id: @m.id
     elsif !@m.published? && (current_user.blank? || !current_user.editor?)
-      flash[:notice] = t(:work_not_available)
+      flash.notice = t(:work_not_available)
       redirect_to '/'
     else
       prep_for_read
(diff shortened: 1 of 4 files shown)


app/controllers/manifestation_controller.rb
```diff
@@ -252,13 +252,11 @@ def dict_entry
   end
 
   def read
-    @m = Manifestation.joins(:expression).includes(:expression).find(params[:id].to_i)
-    if @m.nil?
```
damisul: removed ineffective check (@m cannot be nil as in this case RecordNotFound is thrown)

app/controllers/manifestation_controller.rb
```diff
@@ -252,13 +252,11 @@ def dict_entry
...
-      head :not_found
-    elsif @m.expression.work.genre == 'lexicon' && DictionaryEntry.where(manifestation_id: @m.id).count > 0
+    @m = Manifestation.find(params[:id].to_i)
+    if @m.expression.work.genre == 'lexicon' && DictionaryEntry.exists?(manifestation: @m)
```
damisul: `exists?` check is more performant compared to `.count > 0`

app/controllers/manifestation_controller.rb
```diff
@@ -267,14 +265,14 @@ def read
       @tagging = Tagging.new
       @tagging.taggable = @m
       @taggings = @m.taggings
-      @recommendations = @m.recommendations
-      @my_pending_recs = @recommendations.all_pending.where(user: current_user)
-      @app_recs = @recommendations.all_approved
```
damisul: Second SQL query to fetch approved recommendations

app/views/manifestation/_metadata.html.haml
```diff
@@ -70,6 +70,6 @@
         %span{:style => "font-weight:bold"}= t(:tags)+': '
         = render partial: 'shared/taggings', locals: {taggings: @taggings}
     .metadata.mobile-work-details
-      - if @app_recs.count > 0
+      - if @app_recs.present?
         %span{:style => "font-weight:bold"}= t(:recommendations)+': '
-        %span.pointer.linkcolor#jump-to-recs= t(:this_work_has_x_recommendations, x: @recommendations.all_approved.count)
```
damisul: and one more SQL count query here (probably also should be cached)

damisul (commented):
Hi @abartov , I've did not found big problems related to SQL but did some small cleanup. It saved about 0.2 seconds of read page load time on my machine.

PR #731, opened 2025-11-18
Manifestations caching

damisul:
This PR caches HTML produced from markdown and rendered in manifestation#read action.

It also extracts some logic in three services:
- MakeHeadingIdsUnique
- SanitizeHeadings
- ManifestationHtmlWithChapters

All of them are pure ruby operation, but on a large files it can produce high CPU usage, as it involves massive text manipulations and regexp matching.

### DIFF
--- a/config/locales/he.yml
+++ b/config/locales/he.yml
@@ -1059,7 +1059,7 @@ he:
   record_source: מקור הרשומה
   select_person_first: יש לשייך יוצר/ת לפני שיוך פריטים
   select_publication_first: יש ללחוץ על שורת פריט משמאל לפני שיוך עותק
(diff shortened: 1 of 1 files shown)


lib/bybe_utils.rb
```diff
@@ -777,16 +777,6 @@ def replace_with_redirect(m_id_to_delete, m_id_to_redirect_to)
     h.save
   end
 
-  def make_heading_ids_unique(html)
```
damisul: This method duplicated method in ApplicationController and was used outside of controller context. Now it is in service so we can safely remove it.

abartov (approved):

PR #732, opened 2025-11-19
Speeding up Author#TOC page

damisul:
This PR includes several changes to make TOC page faster:

- Added preloading of data required for rendering TOC to GenerateTocTree service.
- Optimized some lines in _toc_node.html.haml to get benefit from preloaded data.
- Converted Collection.parent_collection_items method to relation so it could be preloaded

### DIFF
--- a/app/views/authors/_toc_node.html.haml
+++ b/app/views/authors/_toc_node.html.haml
@@ -57,8 +57,7 @@
     - collection = toc_node.collection
     - type_prefix = collection.collection_type == 'uncollected' ? '' : textify_collection_type(collection.collection_type)
     - uncollected = collection.collection_type == 'uncollected'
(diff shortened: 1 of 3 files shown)


app/views/authors/_toc_node.html.haml
```diff
@@ -57,8 +57,7 @@
     - collection = toc_node.collection
     - type_prefix = collection.collection_type == 'uncollected' ? '' : textify_collection_type(collection.collection_type)
     - uncollected = collection.collection_type == 'uncollected'
-    - any_published = collection.any_published_manifestations?
```
damisul: this variable was not used, but it caused additional SQL query, so I've removed it

abartov (approved):

PR #745, opened 2025-11-24
Collection show tweaks

damisul:
This PR add preloading of nested and parent collections and their data used to render collections show page.
As result rendering of https://benyehuda.org/collections/10 page on my machine takes **only 94 SQL queries instead of 1663** before.

It also contains several fixed not directly related to performance:
- removed `rails-footnotes` gem as it caused errors in dev environment
- fixed bug, when collections containing no manifestations (e.g. only placeholders) was not viewable due to `no such item` error
- removed some unused actions and methods left after previous refactorings

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -137,7 +137,6 @@ group :development do
   gem 'pronto-haml', require: false
   gem 'pronto-rubocop', require: false
   gem 'rails-erd' # for making ER diagrams
-  gem 'rails-footnotes'
   gem 'rubocop', '1.79.1', require: false
   gem 'rubocop-factory_bot', require: false
   gem 'rubocop-rails', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -584,8 +584,6 @@ GEM
       activesupport (>= 4.2)
       choice (~> 0.2.0)
       ruby-graphviz (~> 1.2)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -137,7 +137,6 @@ group :development do
   gem 'pronto-haml', require: false
   gem 'pronto-rubocop', require: false
   gem 'rails-erd' # for making ER diagrams
-  gem 'rails-footnotes'
```
damisul: removed failing gem

app/controllers/collections_controller.rb
```diff
@@ -8,31 +8,23 @@ class CollectionsController < ApplicationController
...
-    @collections = Collection.all
-  end
+  before_action :set_collection, only: %i(show update destroy)
 
   # GET /collections/1 or /collections/1.json
   def show
-    unless @collection.has_multiple_manifestations?
-      ci = @collection.flatten_items.select { |x| x.item_type == 'Manifestation' }.first
-      if ci.nil?
-        flash[:error] = t(:no_such_item)
```
damisul: We've discussed this over Telegaram and agreed it was a bug: if there were no manifestations in collection (including subcollections) then 'no such item' message was shown.

PR #753, opened 2025-11-26
Added data preloading to Author#latest_popup action to avoid N+1 problem

damisul:
also refactored method to use ruby group_by instead of custom code

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,7 @@
 BY_LANGUAGE=he # locale used for UI
 ELASTICSEARCH_HOST=127.0.0.1:19200
 REDIS_URL=redis://127.0.0.1:16379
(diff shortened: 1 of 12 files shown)


app/controllers/authors_controller.rb
```diff
@@ -80,14 +80,8 @@ def whatsnew_popup
 
   def latest_popup
     @pubs = Rails.cache.fetch("au_#{@author.id}_latest_popup", expires_in: 12.hours) do
-      pubs = @author.latest_stuff
-      pubscoll = {}
-      pubs.each do |m|
```
damisul: replaced this block with ruby's `group_by`

PR #755, opened 2025-11-26
Welcome page speed improvements

damisul:
This PR resolves couple of issues related to N+1 problems on welcome page.
This issue was not to high-priority as they were cached, but I've resolved them anyway:

- added preloading of persons object in block used to display popular authors (person is required to get gender)
- added preloading of involved authorities for popular and most recent works

Also we had too many caching layers for work lists, so I've removed some excessive caching and only left caching on page level.

Also this PR updates development environment to use mem_cached server defined in  docker-compose (previously it expected memcached  to be available on localhost with default port 11211)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -21,7 +21,7 @@ Rails.application.configure do
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :mem_cache_store, '127.0.0.1', {namespace: ENV['CACHE_NONCE'], value_max_bytes: 40*1024*1024}
+    config.cache_store = :mem_cache_store, { namespace: ENV['CACHE_NONCE'], value_max_bytes: 40*1024*1024 }
     #config.cache_store = :memory_store
     config.public_file_server.headers = {
       'Cache-Control' => "public, max-age=#{2.days.to_i}"
--- a/.env
+++ b/.env
@@ -1,3 +1,7 @@
 BY_LANGUAGE=he # locale used for UI
 ELASTICSEARCH_HOST=127.0.0.1:19200
 REDIS_URL=redis://127.0.0.1:16379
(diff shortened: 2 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -101,10 +101,6 @@ def require_crowdsourcer
     redirect_to '/', flash: { error: I18n.t(:not_a_crowdsourcer) }
   end
 
-  def popular_works
```
damisul: removed this wrapper, as it was used only in a one place

app/views/welcome/_works_list.html.haml
```diff
@@ -1,6 +1,5 @@
-- cache cache_key, expires_in: 24.hours do
```
damisul: caching from _works_list partial is removed, now we cache popular and new works as a single fragment in _homepage partial

config/environments/development.rb
```diff
@@ -21,7 +21,7 @@
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :mem_cache_store, '127.0.0.1', {namespace: ENV['CACHE_NONCE'], value_max_bytes: 40*1024*1024}
```
damisul: removed hardcoded memcached address, so Rails will use value from MEMCACHE_SERVERS env variable instead

config/initializers/scheduler.rb
```diff
@@ -11,9 +11,6 @@
   #puts "calculating recommendation counts..."
   #Person.recalc_recommendation_counts
 end
-scheduler.every '36h' do
-  Manifestation.get_popular_works
```
damisul: I believe the purpose of this was to prepopulate Popular works cache by schedule, but I think it is redundant.

abartov (approved):

PR #759, opened 2025-11-27
Newsfeed fix

damisul:
This PR reverts back changes we did during recent production failures.

- turns on newsfeed again (NOTE: see my comment in telegram)
- returns back PR #755 
- removes index from ahoy_events table causing problem (migration uses `if exists` syntax so it should not fail if index already deleted manually)
- changed Newsfeed caching: instead of caching NewItems (we were getting  empty authority_id after restoring values from MemCached) it now caches whole newsfeed html snippet. It required to extract some logic into a service for easier call from HAML code.

### DIFF
--- a/app/controllers/concerns/tracking.rb
+++ b/app/controllers/concerns/tracking.rb
@@ -27,21 +27,19 @@ module Tracking
   def track_view(record)
     return if spider?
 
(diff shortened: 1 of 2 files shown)


abartov:
Is this ready?

abartov (approved):

PR #801, opened 2025-12-18
Add gender enum field to LexPerson

abartov:
## Summary
- Added gender enum field to LexPerson model (mirroring Person model)
- Gender options: male, female, other, unknown
- Default value: male (0)
- Added gender select field to editing form
- Added gender_letter method for Hebrew text formatting
- Updated views to use gendered Hebrew text

## Implementation Details
- Created migration to add gender:integer column with default 0
- Added enum definition to LexPerson model
- Updated controller to accept gender parameter
- Added gender select in form between copyrighted and birthdate fields
- Fixed migration idempotency issues in two unrelated migrations

## Test Plan
- [x] Ran existing tests - all passing
- [x] Verified migration runs successfully
- [x] Verified enum works correctly in Rails console
- [x] Default gender is set to male

Implements by-n1i

### DIFF
--- a/app/models/lex_person.rb
+++ b/app/models/lex_person.rb
@@ -18,7 +18,7 @@ class LexPerson < ApplicationRecord
   end
 
   def gender_letter
-    return gender == 'female' ? 'ה' : 'ו'
+    female? ? 'ה' : 'ו'
   end
 
   def self.find_or_create_by_authority_id(aid)


abartov:
@damisul - FYI, had to add it for display purposes, as Hebrew is a gendered language, and it was impossible to properly present LexPersons without being able to set their gender (and while it could be drawn from Person if linked, a significant portion of LexPersons won't be linked to Persons).

PR #806, opened 2025-12-18
Fix flaky collection search tests

abartov:
## Summary
Fixed 13 failing tests in `search_collections_spec.rb` and `collections_controller_spec.rb` by addressing two root causes identified through git bisect analysis.

### 1. Factory Creating Non-Indexable Collection Types
The collection factory randomly selected from 5 types including `periodical_issue`:
```ruby
collection_type { %w(volume periodical periodical_issue series other).sample }
```

But `CollectionsIndex` excludes this type from indexing:
```ruby
index_scope Collection.where.not(collection_type: %w(periodical_issue uncollected))
```

This caused ~20% of test collections to not be indexed, leading to random test failures.

### 2. Controller Tests Using Eager Loading
Controller tests used `let!` (eager loading) which created records BEFORE the `Chewy.strategy(:atomic)` block. The atomic strategy only indexes records created INSIDE the block, so eagerly-created records were never indexed.

## Changes
- `spec/factories/collections.rb`: Removed `periodical_issue` from random sample
- `spec/controllers/collections_controller_spec.rb`: Changed `let!` to `let` for browse test collections

## Test Results
- Before: 51 examples, 13 failures
- After: 51 examples, 0 failures ✅

## Related
Closes by-tho


abartov:
Closing to recreate with correct base branch (lexicon_new)

PR #838, opened 2025-12-25
Updated rmultimarkdown gem to v6.7

damisul:
After upgrade of my machine from Debian Bookworm to Debian Trixie it turned out rmultimarkdown gem cannot be compiled there. we need to add either macro define in one of C files, or add special flag to compiler.

I've contacted upstream author and as the first step he updated binary dependency from MultiMarkdown_6.4 to 6.7, but it did not worked.
So there is a PR to upstream adding flag to compiler https://github.com/tillsc/multi_markdown/pull/10, but it is not accepted yet.
So I've switched project to use my fork for now.

This version compiles well both in old and new Debian systems.
But as undelying library was updated as well, it changed behaviour a little bit and it caused spec failure.
Looks like newer version generates slightly different HTML for footnotes, see changes in affected spec (it adds additional `&#xfe0e;` character to footnotes arrow).

&#xfe0e; - is a special "Variation Selector" character which affects visual appearance of previous character (↩︎)
See https://www.ii.com/unicode-variation-selectors-15-16/ for details

### DIFF
--- a/app/controllers/anthologies_controller.rb
+++ b/app/controllers/anthologies_controller.rb
@@ -255,6 +255,14 @@ class AnthologiesController < ApplicationController
 
   # PATCH/PUT /anthologies/1
   def update
+    # Prevent non-admin users from setting access to 'pub'
+    if !current_user&.admin? && anthology_params[:access] == 'pub'
+      @anthology.errors.add(:access, I18n.t('errors.messages.not_authorized',
+                                            default: 'You do not have permission to set access to public.'))
+      respond_with_error
+      return
+    end
+
     @anthology.update!(anthology_params)
     @cur_anth_id = @anthology.nil? ? 0 : @anthology.id
     @error = false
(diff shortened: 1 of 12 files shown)


docker/Dockerfile.base
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:24.04
 ADD https://github.com/jgm/pandoc/releases/download/3.8.2.1/pandoc-3.8.2.1-1-amd64.deb /tmp/pandoc.deb
 
 RUN apt-get update -qq \
-    && apt-get install -y curl cmake wkhtmltopdf yaz libyaz-dev libmagickwand-dev \
+    && apt-get install -y curl cmake git wkhtmltopdf yaz libyaz-dev libmagickwand-dev \
```
damisul: I had to add git to docker image used to run specs on Github, as we use github source in gemfile now

spec/services/manifestation_html_with_chapters_spec.rb
```diff
@@ -206,11 +206,11 @@
           <ol>
 
           <li id="fn:1">
-          <span>Footnote 1 <a href="#fnref:1" title="return to body" class="reversefootnote">&#160;&#8617;</a></span>
+          <span>Footnote 1 <a href="#fnref:1" title="return to body" class="reversefootnote">&#160;&#8617;&#xfe0e;</a></span>
```
damisul: Variation Selector invisible character added by newer version of MultiMarkdown

PR #960, opened 2026-01-29
Optimize system tests

damisul:
- Used shorter notation in all system tests (`desribe` instead of `Rspec.describe`)
- Used webdriver with javascript only where neccessary (greately reduces time of non-js dependent test)
- Simplified capybara config (we will use only Firefox as the js driver)
- Fixed few bugs in existing specs (there are still 6 failing tests though)


spec/support/capybara.rb
```diff
@@ -9,43 +9,18 @@
...
-               :selenium_firefox_headless
-             else
-               :selenium_chrome_headless
-             end
-    driven_by driver
-  rescue StandardError => e
-    # If both browsers fail, skip JS tests with a warning
-    skip "Skipping JS test - Browser/Selenium not properly configured: #{e.message}"
+    driven_by :selenium_firefox_headless
+    page.driver.browser.manage.window.resize_to(1400, 900)
```
damisul: Main fix.
Added page resize before each test. It seems the problem was that at some point browser window was minimized and mobile view was rendered.

spec/system/admin/tags_management_spec.rb
```diff
@@ -2,11 +2,7 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Admin Tags Management', :js, type: :system do
-  before do
-    skip 'WebDriver not available or misconfigured' unless webdriver_available?
```
damisul: Removed such checks here and below in file. Test environment should be in a working state everywhere. In CI we've disable system tests so it should be OK too

spec/system/lexicon/verification_workbench_spec.rb
```diff
@@ -460,19 +428,22 @@ def vips_available?
 
   describe 'Profile Image Selection', :js do
     before do
-      skip 'Vips library not available for image processing' unless vips_available?
-
       # Attach some test images to the entry
-      entry.attachments.attach(
-        io: StringIO.new('test image 1'),
```
damisul: Such attachments caused errors, as ActiveStorage tried to analyze jpeg image, but it had wrong format. Replaced it with real png images (added them to fixtures folder)

Issue #972, opened 2026-02-03
Create LexEntries for every Lexicon entry from the outset


abartov:
I just realized that in order to give readers the full Lexicon experience in the months of interim migration, we need to create LexEntry entities for **all** the entries in the legacy Lexicon, not just the LexPerson ones we can handle so far. Otherwise they wouldn't show in the /lex/list view.

Those unmigrated and for-now-unmigratable LexEntries would simply redirect to the legacy URLs with the PHP files until we do migrate them.

This would require some care, as there may already be some assumptions about LexEntries.

damisul:
Hi @abartov , I've did first iteration on this work: https://github.com/projectbenyehuda/bybe/commit/4e5f2c59287ac22bc1fb05cfb01175faef7abaa3

It is already in a lexicon branch.

Actually we already create LexEntries for all PHP files. It was required to properly update links in markdown with a links to a new entries.

So I've updated Lexicon::Entries controller to
- display all not yet migrated entries in a #list action
- updated #show action to redirect to old lexicon if entry is not yet migrated

This is work in progress, but I believe it will unblock you.
I'll continue working on it, but please review if it is correct direction?

abartov:
Now that I'm seeing the full lexicon in /lex/list, I noticed the English and miscellaneous stuff is appearing before the Hebrew (due to naive alphabetical sort by character code...). We want the Hebrew stuff to be first, and anything not starting with a Hebrew letter or a digit to be relegated toward the end of the list. 

I think a simple implementation could be to:
1. add a **sort_title** field (indexed)
2. check the title _during ingestion_, and if it doesn't start with Hebrew/digit (you can use my 'hebrew' gem to check for Hebrew), populate the sort_title with something like "תתתת_#{title}" (those four characters are the last letter in the Hebrew alphabet, and won't occur in such a sequence in the data (like 'ZZZZ'), guaranteeing non-Hebrew goes to the bottom of the list.
3. Display entries in /lex/list in sort_title order.

What do you think?

PR #1001, opened 2026-02-15
Extracted logic for sort_title generation to SortedTitle concern

damisul:
See https://github.com/projectbenyehuda/bybe/issues/972#issuecomment-3854200728

This PR updates logic used to generate sort_title value in Manifestation and Collection:
1) I've removed method from ApplicationRecord class and extracted it into new SortedTitle Concern (and plugged this concern into Manifestation and Collection classes). Just better code organization.
2) I've moved autogeneration logic from before_save to before_validation (usually it is better to have normalization done before validation)
3) I've updated sort_title generation to add special prefix to titles started nor from Hebrew, nor from Digit (i.e. mostly for Latin-based titles) so such titles will be sorted after all Hebrew titles.
4) Added specs for it

### DIFF
--- a/app/assets/javascripts/ingestibles_collection_selector.js
+++ b/app/assets/javascripts/ingestibles_collection_selector.js
@@ -1,0 +1,119 @@
+// Collection selector and sub-collection functionality for Ingestibles#edit form
+
+$(document).ready(function() {
+  // Override the autocomplete source to include collection_type filter
+  $('#cterm').on('railsAutocomplete.source', function(event, data) {
+    var collectionType = $('#collection_type_filter').val();
+    if (collectionType && collectionType !== 'all') {
+      data.extraParams = { collection_type: collectionType };
+    }
+  });
+
+  // Reload autocomplete when collection type filter changes
+  $('#collection_type_filter').on('change', function() {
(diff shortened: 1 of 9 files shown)


app/models/concerns/sorted_title.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Removing special symbols (quotes, parenthesis, hyphens, etc.) from sort_title
+    self.sort_title = title.strip_nikkud.tr('[]()*"\'', '').tr('-־', ' ').strip
+
+    # If sort title starts with number and dot, e.g. "12. Title" we remove numeric prefix and leave only "Title"
+    self.sort_title = ::Regexp.last_match.post_match if sort_title =~ /^\d+\. /
+
+    # If sort_title starts from non-hebrew characters we add a prefix to it to make it appear after all hebrew titles
+    unless sort_title.blank? || sort_title[0].any_hebrew? || sort_title.start_with?(/\d/)
+      # those four characters are the last letter in the Hebrew alphabet
+      self.sort_title = "תתתת_#{sort_title}"
```
damisul: Main change in logic

PR #1024, opened 2026-02-24
Added section to describe our approach to i18n

damisul:
I've added some instructions for I18n, hopefully it will help AI to generate better i18n resources


abartov (changes requested):
This is great, thank you!

It would be more effective in AGENTS.md, where agents are more likely to read it. I don't think the README.md is read unless explicitly pointed to.

damisul:
> This is great, thank you!
> 
> It would be more effective in AGENTS.md, where agents are more likely to read it. I don't think the README.md is read unless explicitly pointed to.

Ok, you're right. I've extracted i18n guide in a separate file and referenced it in AGENTS.md. So it can be read both by human developers and by AI agents.

damisul:
@abartov , is updated version OK?

abartov (approved):

PR #1168, opened 2026-04-21
Fix URI.open vulnerability in IngestiblesController

abartov:
## Summary

- Replace `URI.open(params[:docx])` with `Faraday.get` (already used in the project) to eliminate the `Kernel.open` shell-injection risk
- Add an explicit `http`/`https` scheme guard that raises `ArgumentError` before the HTTP request for any non-http(s) URL (e.g. `file://`)
- Pipe-prefixed strings like `|rm -rf /` are rejected even earlier by `URI.parse` raising `URI::InvalidURIError`

## Security context

`URI.open` (from `open-uri`) ultimately delegates to `Kernel.open`. A string starting with `|` causes `Kernel.open` to execute the remainder as a shell command. Since `params[:docx]` is user-controlled, this path allowed remote code execution.

## Test plan

- [x] New `when docx URL param is provided` context in `spec/controllers/ingestibles_controller_spec.rb` covering:
  - Valid `https://` URL: Faraday is called, file is attached with correct filename
  - Pipe-prefixed string (`|rm -rf /`): `URI::InvalidURIError` raised, Faraday never called
  - `file://` URL: `ArgumentError` raised, Faraday never called
- [x] Full controller spec (58 examples, 0 failures)

Fixes beads issue beads_by-8g9.

### DIFF
--- a/app/controllers/ingestibles_controller.rb
+++ b/app/controllers/ingestibles_controller.rb
@@ -156,9 +156,11 @@ class IngestiblesController < ApplicationController
       end
     end
     if params[:docx].present?
-      file = URI.open(params[:docx])
       uri = URI.parse(params[:docx])
-      @ingestible.docx.attach(io: file, filename: CGI.unescape(File.basename(uri.path)))
+      raise ArgumentError, 'Only http/https URLs are allowed' unless %w(http https).include?(uri.scheme)
+
+      response = Faraday.get(params[:docx])
+      @ingestible.docx.attach(io: StringIO.new(response.body), filename: CGI.unescape(File.basename(uri.path)))
     end
     if @ingestible.save
       redirect_to edit_ingestible_url(@ingestible), notice: t('.success')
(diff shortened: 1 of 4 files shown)


abartov (commented):
Addressed all 5 comments in commit 32611c386:

**Comments 3, 4, 5 (error handling / exception bubbling):** The `params[:docx]` block now rescues `URI::InvalidURIError` and `ArgumentError` (bad scheme) into a 422 + `@ingestible.errors.add(:base, ...)` + `render :new`. `Faraday::Error` is also rescued the same way, and `response.success?` is checked before attaching. Specs were updated to assert HTTP 422 status and non-empty `errors[:base]` instead of raised exceptions. Added a new spec for the non-success Faraday response case.

**Comment 2 (memory / StringIO):** Switched from `StringIO.new(response.body)` to `Tempfile.new(...)` — consistent with how `HtmlFileController` handles the same pattern.

**Comment 1 (SSRF):** Agreed this is a real concern. It's a pre-existing architectural issue (`skip_before_action :verify_authenticity_token` was there before this PR and is needed for the task-system integration). Tracked as a follow-up in beads_by-wj7 to evaluate host allowlisting / private-IP blocking / token auth for that code path.

PR #1263, opened 2026-05-26
Allow re-running migration from escalated status

abartov:
## Summary

- `escalated` entries had no redo-migrate button in `/lex/files`, leaving them stuck just like the `draft`/`verifying` + nil-lex_item case fixed in #1262
- Controller `redo_migration` guard now accepts `escalated` alongside `draft` and `verifying`
- View: extracted the three permitted statuses into a `redo_eligible` local variable, eliminating duplication between the `lex_item.present?` inner branch and the nil-lex_item fallback, and fixing the haml-lint line-length violation that the repeated condition caused

## Test plan

- [x] New spec context: `escalated` entry triggers redo-migration job and transitions to `migrating`
- [x] Boundary spec updated: excludes `escalated` from the "should not queue" path
- [x] All 15 examples pass, haml-lint and rubocop clean


abartov:
Closing to recreate cleanly on top of the rebased lexicon branch.

PR #1298, opened 2026-06-03
Report unreachable external links as broken in lexicon verification

abartov:
## Problem

In the lexicon migration verification workbench, a defunct external link such as `http://www.mouse.co.il/CM.articles_item,1045,209,52641,.aspx` (whose host no longer resolves) was **not** reported as inaccessible.

### Root cause

`Lexicon::CheckExternalLinks` stores `nil` in `link_http_status` whenever a check fails to produce an HTTP code — DNS failure, connection refused, timeout, or SSRF-blocked host. But the predicates that drive the UI badge only flagged a link when the status was a present integer `>= 400`:

```ruby
def link_broken? = link_http_status.present? && link_http_status >= 400
```

So a **totally dead host** (the most inaccessible case) was silently treated as healthy, while only servers alive enough to answer with a 4xx/5xx were flagged. Worse, `nil` conflated two distinct states: "never checked" and "checked but unreachable."

## Fix

Add a `checked_at` timestamp (`link_checked_at` on `lex_citations`, `checked_at` on `lex_links`) to disambiguate, and redefine broken-ness:

```ruby
broken? = checked_at.present? && (http_status.nil? || http_status >= 400)
```

- **Migration** backfills `checked_at = updated_at` for rows that already have a stored status, preserving existing 4xx flagging. Rows with a `nil` status (ambiguous) are left `NULL` and get an authoritative result on the next check run.
- `CheckExternalLinks` and the interactive citation recheck now stamp `checked_at` on every check.
- Recheck of an unreachable link now shows a red **error** toast ("inaccessible") instead of the amber "network error" warning.
- New `broken_link_badge` helper renders an "unreachable" badge (no HTTP number) when the status is nil; numeric statuses render as before.
- New i18n key `broken_link.unreachable` (he + en); removed the now-unused `check_failed`.

## Tests

- **Model specs** (`lex_citation_spec`, new `lex_link_spec`): table-driven cases for `link_broken?`/`broken?` across never-checked / unreachable / 200 / 404 / 500.
- **Service spec**: network-failure and DNS-failure cases now assert the link is flagged broken and `checked_at` is recorded.
- **Request spec**: unreachable recheck stores nil status + sets `checked_at` + flags broken + returns an error toast; link-clear resets both columns.
- **System spec**: unreachable link now shows the broken-link badge and a red error toast.

All affected specs pass (62 model/service/request examples + 7 system examples). Rubocop/haml-lint: no new offenses introduced (pre-existing file-wide style lints left untouched).

Fixes beads_by-0p0.

### DIFF
--- a/app/controllers/lexicon/citations_controller.rb
+++ b/app/controllers/lexicon/citations_controller.rb
@@ -130,7 +130,7 @@ module Lexicon
 
     def link_toast_for(status)
       if status.nil?
-        ['error', t('lexicon.verification.broken_link.unreachable')]
+        ['error', t('lexicon.verification.broken_link.inaccessible')]
       elsif status < 400
         ['success', t('lexicon.verification.broken_link.now_accessible')]
       else
--- a/app/controllers/lexicon/citations_controller.rb
+++ b/app/controllers/lexicon/citations_controller.rb
@@ -114,12 +114,12 @@ module Lexicon
 
     def check_link_synchronously
       if @citation.link.blank?
-        @citation.update_column(:link_http_status, nil)
+        @citation.update_columns(link_http_status: nil, link_checked_at: nil)
         return
       end
 
       status = CheckExternalLinks.new.check_url(@citation.link)
-      @citation.update_column(:link_http_status, status)
+      @citation.update_columns(link_http_status: status, link_checked_at: Time.current)
       @link_check_performed = true
       @link_toast_type, @link_toast_message = link_toast_for(status)
       # rubocop:disable Rails/ActionControllerFlashBeforeRender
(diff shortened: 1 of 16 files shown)


abartov:
Addressed the Copilot review (commit d57badc5f). Both comments flagged that the nil-status message was too specific — `fetch_status` returns nil for invalid URLs, non-HTTP schemes, SSRF-blocked addresses, and redirect loops too, not just an unreachable host. Renamed the i18n key `broken_link.unreachable` → `broken_link.inaccessible` and reworded to "Inaccessible link (could not be retrieved)" (Hebrew: "קישור לא נגיש (לא ניתן לאחזר אותו)"), which is accurate for every nil cause. Updated the helper, controller, and specs accordingly.

PR #1305, opened 2026-06-05
Preserve person links in plain (non-role) work comments during migration

abartov:
## Problem

During lexicon migration, person links embedded in **plain** work comments were silently dropped. `Lexicon::ParsePersonWork` only recovered links from comments matching a role prefix (`עריכה`/`איורים`/`בשיתוף`/`על`); any other comment fell through to the `plain_comments` branch and was stored as plain text via `element.text`, discarding the `<a>` href.

**Concrete case** (reported): the `מעיל קטון` work in `00022.php` has the comment `כולל אחרית דבר מאת <a href="00032.php">יגאל שוורץ</a>`. `כולל` is not a role prefix, so the link to יגאל שוורץ was lost while the name survived as text.

## Fix

Mirror the existing `title_links` mechanism with a new `comment_links` JSON column:

- **Migration/parse** — `ParsePersonWork#extract_comment_links` collects person anchors inside comment-section (`<font size="2">`) wrappers whose text appears in a plain comment, and stores them as `[{ 'text' =>, 'entry_id' => }]`. Restricting to `font[size="2"]` anchors uses the same convention `extract_title_links` relies on, avoiding false positives from title links.
- **Rendering** — new `render_person_work_comment` helper escapes the comment and substitutes matched names with links; wired into `render_person_work` and the verification person-sections view.
- **Editor UI** — add/remove `comment_link` AJAX actions, routes, and a `comment_links.html.haml` panel in the work edit form, mirroring `title_links`.

## Tests

- `parse_person_work_spec`: plain comment with a person link → stored as text + preserved in `comment_links`; role-prefixed comment still becomes a `linked_person` (not duplicated into `comment_links`).
- `lexicon_helper_spec`: `render_person_work_comment` hyperlinks matched names, escapes HTML, handles nil links and missing entries.
- `person_works_spec` (request): GET/add/remove `comment_link` endpoints incl. blank/non-person validation.
- `person_work_comment_links_spec` (system): link renders on the verification page and the existing link shows in the edit modal.

All new specs pass (78 examples). Pre-existing full-suite failures are an environment artifact (disk space → Elasticsearch read-only) unrelated to this change; every affected spec passes in isolation.

### DIFF
--- a/app/controllers/lexicon/person_works_controller.rb
+++ b/app/controllers/lexicon/person_works_controller.rb
@@ -87,12 +87,9 @@ module Lexicon
     end
 
     def remove_title_link
-      if params[:index].blank?
-        head :unprocessable_content
-        return
-      end
+      index = link_index_param
+      return head :unprocessable_content if index.nil?
 
-      index = params[:index].to_i
       links = Array(@work.title_links)
       links.delete_at(index) if index >= 0 && index < links.size
       @work.update!(title_links: links.presence)
@@ -124,12 +121,9 @@ module Lexicon
     end
 
     def remove_comment_link
-      if params[:index].blank?
-        head :unprocessable_content
-        return
-      end
+      index = link_index_param
+      return head :unprocessable_content if index.nil?
 
-      index = params[:index].to_i
       links = Array(@work.comment_links)
       links.delete_at(index) if index >= 0 && index < links.size
       @work.update!(comment_links: links.presence)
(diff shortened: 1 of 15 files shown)


abartov:
Addressed both Copilot review comments in c93e0ed51:

1. **N+1 in `render_person_work_comment`** — now preloads all referenced `LexEntry` records in a single `where(id: …).index_by(&:id)` and looks them up from the hash, instead of a per-link `find_by`.
2. **`remove_comment_link` index coercion** — `:index` is now parsed via `Integer(params[:index], exception: false)` and the action returns 422 for missing/non-integer input, so a value like `'abc'` can no longer silently delete the first link. Applied the same guard to the sibling `remove_title_link` (same latent bug) for consistency, with regression specs for both.

PR #1311, opened 2026-06-15
Fix items removal

### LINKED ISSUE #1310: Collection-transplant bug still occurs (after collection removal)
@abartov wrote in Telegram:
> BTW, I *think* I finally found a reproducible scenario for the remaining collection-transplant bug:
>
> 0. Create a collection with items A, B, C, D, E.
> 1. Then, load/refresh the collection management view, and do the following without refreshing:
> 2. Create a sub-collection inside it.
> 3. Drag the new sub-collection somewhere above the bottom.
> 4. *Delete* (i.e. remove from the collection) the item below the sub-collection. (<— I think this is crucial)
> 5. Drag the next item under the sub-collection into the empty sub-collection.
>
>This should result in an "Item index mismatch in TransplantItem" error, expected N but got N+1.
>(and the transplant is not performed.)

damisul:
This PR fixes couple of errors, causing 'Wrong Index' errors on manage collection page.

1. When we pressed 'Delete' button on collection (to remove Collection itself) it returned javascript callback which removed collection-related items from page. But! It must also delete html element which is considered by Sortable.js as 'draggable item' and this is not the Collection wrapped, but CollectionItem wrapper. As CollectionItem wrapper stayed on page, Sortable.js assumed this element is still present and 'allocated' index position for it. As result all items below removed Collection got wrong index.

2. Very similar error, but for CollectionItem removal ('remove' link on a left part of item). It also returned Js callback, but this callback did not removed item. Instead it called 'fadeOut' method on it. FadeOut only makes item invisible, but it stays on the page. As result all item below it also got index increased by 1. I've updated callback to actually remove item after fadeOut complete.

P.S. I've ignored linter warnings in haml file to keep PR short

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -255,7 +255,8 @@ class CollectionsController < ApplicationController
   # DELETE /collections/1 or /collections/1.json
   def destroy
     @destroyed_id = @collection.id
-    @collection.destroy
+    @collection_item_ids = @collection.parent_collection_items.pluck(:id)
+    @collection.destroy!
 
     respond_to do |format|
       format.html { redirect_to collections_url, notice: t(:deleted_successfully) }
--- a/app/views/collection_items/destroy.js.erb
+++ b/app/views/collection_items/destroy.js.erb
@@ -1,3 +1,3 @@
-ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
+let ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
 ele.find('.by-card-v02').css('background-color', 'orangered');
-ele.fadeOut(1500);
+ele.fadeOut(1500, function() { $(this).remove(); });
--- a/app/views/collection_items/destroy.js.erb
+++ b/app/views/collection_items/destroy.js.erb
@@ -1,3 +1,3 @@
-let ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
+ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
 ele.find('.by-card-v02').css('background-color', 'orangered');
 ele.fadeOut(1500, function() { $(this).remove(); });
--- a/app/views/collections/destroy.js.erb
+++ b/app/views/collections/destroy.js.erb
@@ -1,4 +1,3 @@
-$('*[id*=_coll_container_<%= @destroyed_id %>]:visible').each(function() {
-    $(this).remove();
-});
-$('#cwrapper_<%= @destroyed_id %>').remove();
+<%= @collection_item_ids.each do |item_id| %>
+  $('div[id*=_collitem_<%= item_id %>]').remove();
+<% end %>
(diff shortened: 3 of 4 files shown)


app/controllers/collections_controller.rb
```diff
@@ -255,7 +255,8 @@ def update
   # DELETE /collections/1 or /collections/1.json
   def destroy
     @destroyed_id = @collection.id
-    @collection.destroy
+    @collection_item_ids = @collection.parent_collection_items.pluck(:id)
```
damisul, line 258: We need to get list of collection_items where removed collection is an item.

app/controllers/collections_controller.rb
```diff
@@ -255,7 +255,8 @@ def update
...
+    @collection.destroy!
```
damisul, line 259: added `!` to ensure error will arise if deletion failed (destroy will only return false)

app/views/collection_items/destroy.js.erb
```diff
@@ -1,3 +1,3 @@
-ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
+let ele = $('div[id*="<%= 'collitem_'+@deleted_id.to_s %>"]');
 ele.find('.by-card-v02').css('background-color', 'orangered');
-ele.fadeOut(1500);
+ele.fadeOut(1500, function() { $(this).remove(); });
```
damisul, line 3: remove html element after fade-out

app/views/shared/_editable_collection.html.haml
```diff
@@ -13,73 +13,72 @@
           %b.white= '⇇'
         .collection_focus_button.by-button-v02{ title: t('.focus_on_this_collection') }
           %b= link_to '⤝', collection_manage_path(collection.id), class: 'white'
-  - if collection
```
damisul, line 16: this `if` is redundant. We already used `collection` above, so if it were nil we'd got an exception

app/views/collections/destroy.js.erb
```diff
@@ -1,4 +1,3 @@
-$('*[id*=_coll_container_<%= @destroyed_id %>]:visible').each(function() {
-    $(this).remove();
-});
-$('#cwrapper_<%= @destroyed_id %>').remove();
\ No newline at end of file
+<%= @collection_item_ids.each do |item_id| %>
+  $('div[id*=_collitem_<%= item_id %>]').remove();
```
damisul, line 2: we must remove html element representing 'draggable item', i.e. CollectionItem wrapper.

abartov (changes requested):
Removing collection items in manage_toc does remove them, but the display does not update. Only a page refresh does.

abartov:
For example, http://localhost:3000/author/391/manage_toc should give you a lot of complex collections to play with.

damisul:
> Removing collection items in manage_toc does remove them, but the display does not update. Only a page refresh does.

@abartov , my bad. I've added `let` keyword in js callback to properly declare variable, but it seems this approach does not works in js views: it works fine on first execution, but on a second one it fails with `Uncaught SyntaxError: redeclaration of let ele`.

So I've removed `let`, it should works now.

PR #1450, opened 2026-07-10
Author TOC: alphabetical Collections sort keeps one-card-per-collection layout

abartov:
## Problem

On the Authority TOC, the **Titles: Alphabetical** sort (`colls_abc`) shared the
flat-sort rendering tail, so it dumped every collection into a **single
mega-card** instead of the one-card-per-collection layout used by the default
chronological (`colls`) sort. In addition, the uncollected-works collection —
whose collection title is empty — sorted to the **very top** of the alphabetical
list.

## Fix

`colls_abc` now gets its own branch in the `#sort_by` handler (in
`app/views/authors/_generated_toc.html.haml`):

- Restores the chronological, server-rendered one-card-per-collection layout
  (role + section headers intact).
- Reorders the collection cards **within each section list**
  (`ul.toclist[role="tree"]`) alphabetically by title, via a new
  `sortCollectionsAbc()` helper. Each collection is moved together with its
  empty wrapper `<li>` and trailing `<br>` so per-card spacing is preserved.
- Uncollected works keep their own trailing section (empty title sorts last), so
  they appear at the **end** instead of the beginning.

The flat manifestation-level sorts (title/popularity/date) are unchanged — they
were just promoted out of the now-removed inner `if/else`.

## Layout choice

Matches the chronological view's structure exactly (role + section headers),
only reordering cards — per the requested behavior.

## Tests

- New: `spec/system/author_toc_colls_abc_spec.rb`
  - alphabetical reorder keeps one card per collection (no `#sorted_card`, ≥2 treeitem cards)
  - uncollected works stay at the very end, not the beginning
- Regression-checked: `author_toc_filters_spec.rb` (8), `author_toc_volume_collapse_toggle_spec.rb` (4) — all green.

### DIFF
--- a/app/views/authors/_generated_toc.html.haml
+++ b/app/views/authors/_generated_toc.html.haml
@@ -70,6 +70,36 @@
         }
       });
     }
+    // Alphabetical Collections sort ('colls_abc'): keep the chronological
+    // one-card-per-collection layout and only reorder the top-level collection
+    // cards within each section list (ul.toclist[role=tree]) by title.
+    // Each collection is emitted as a trio [empty wrapper <li>, card <li>, <br>];
+    // the trio is moved together so per-card spacing/margins are preserved.
+    // Uncollected works have an empty .ctitle_only and always sort to the end
+    // (in practice they sit alone in their own trailing section).
+    function sortCollectionsAbc() {
+      $('#browse_mainlist ul.toclist[role="tree"]').each(function() {
+        var $list = $(this);
+        var $cards = $list.children('li.by-card-v02');
+        if($cards.length < 2) { return; }
+        var units = $cards.map(function() {
+          var $card = $(this);
+          var $unit = $card;
+          var $prev = $card.prev();
+          if($prev.is('li') && !$prev.hasClass('by-card-v02')) { $unit = $prev.add($unit); }
+          var $next = $card.next();
+          if($next.is('br')) { $unit = $unit.add($next); }
+          return { key: $card.find('.ctitle_only').first().text().trim(), nodes: $unit };
+        }).get();
+        units.sort(function(a, b) {
+          if(!a.key && !b.key) { return 0; }
+          if(!a.key) { return 1; }  // uncollected / untitled last
+          if(!b.key) { return -1; }
+          return a.key.localeCompare(b.key);
+        });
+        units.forEach(function(u) { $list.append(u.nodes); });
+      });
+    }
     $('#sort_by').change(function(){
       var flat = isFlatSort(this.value);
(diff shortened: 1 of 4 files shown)


abartov:
Addressed the review + the reported filter-loss bug by **unifying the two genre-filtering mechanisms** (per follow-up guidance), rather than patching them to coexist.

**Copilot comments**
- _Double \`applyGenreFilters\` call_ / _stale node cache after \`#browse_mainlist\` rebuild_: both are moot now — `applyGenreFilters` (the separate grouped-view genre toggle) is **removed entirely**. The author-card genre chips are now shortcuts that switch into flat/filter mode and toggle the matching filters-pane genre checkbox (`syncGenreChips()` mirrors the pane facet back onto the chips). One mechanism, no cache.
- _Spec comment: `sort_term` is `[normalized_pub_year || created_at.year, id]`_: corrected.

**github-actions (rubocop)**
- `RSpec/ScatteredSetup` / `RSpec/ScatteredLet` in the new spec: merged the `before` hooks and grouped the `let` blocks; also tidied the pre-existing scattered `before` in `author_toc_filters_spec.rb`.

**Reported bug (filter lost on sort change)**
- Root cause was `applyGenreFilters` re-showing every node of an active genre after the pane had filtered — gone with the unified mechanism. Additionally, flat→flat re-sorts now preserve **all** active filters (incl. date range) via `TocFilters.reapply()`.

**New tests**
- Genre-chip-as-shortcut behavior (enter filter mode, toggle genre, toggle off).
- Filter preservation across flat-list sort changes.
All author-TOC system specs green (filters 10, colls_abc 2, volume-collapse 4).

abartov:
Addressed the latest Copilot review (both comments, on the two spec files).

The catch: my earlier merge of the `before` hooks (to satisfy rubocop's `RSpec/ScatteredSetup`) meant the `let!` DB/Chewy setup ran *before* the WebDriver skip could short-circuit. Copilot's suggested fix (a separate skip `before` placed before the `let!`s) would re-trigger `ScatteredSetup`, so I resolved the tension by converting the record declarations from `let!` to **lazy `let`** and keeping a single `before` with the skip first. Now nothing expensive runs until it's referenced in the `before` (after the skip), and there's still only one `before` hook — both concerns satisfied.

- `spec/system/author_toc_colls_abc_spec.rb`: `author`/`zebra`/`apple` → lazy `let` (creation order in the before still forces `zebra` before `apple`, preserving the chronological-vs-alphabetical distinction).
- `spec/system/author_toc_filters_spec.rb`: `author`/`volume`/`poem`/`story` → lazy `let`.

Both specs rubocop-clean; all 12 examples green.

PR #1462, opened 2026-07-15
Kwic refactoring

### LINKED ISSUE #1440: Refactor the way we check if KWIC report is being generated
We schedule KWIC Concordance generation only when it is not already scheduled for given Authority of Collection, see

`GenerateKwicConcordanceJob.in_progress?` method

https://github.com/projectbenyehuda/bybe/blob/0850f81a7e9bc071eb385a870b318bb83477edbc/app/sidekiq/generate_kwic_concordance_job.rb#L15

The problem with it is that it uses low-level Sidekiq API, and it can be changed with major Sidekiq update.
Another issue is that it is very hard to test and debug it. And of course it will not work with any other ActiveJob backend.

I'd propose to use some different approach to resolve it.
E.g. we can add `kwic_generation_scheduled_at` column to Authorities/Collections tables and fill it before scheduling KWIC generation.

damisul:
Updated KWIC generation Job to check if job is already queued without using low-level Sidekiq API.
Added column :kwic_generation_started_at to Authorities and Collections table instead

It also fixes small issue preventing kwic generation to work in docker containers (explicitely set path to temp file)


app/jobs/generate_kwic_concordance_job.rb
```diff
@@ -44,9 +21,8 @@ def perform(entity_type, entity_id)
     end
   rescue ActiveRecord::RecordNotFound => e
     Rails.logger.error("GenerateKwicConcordanceJob: #{entity_type} with id #{entity_id} not found: #{e.message}")
-  rescue StandardError => e
```
damisul, line 47: this should be managed by ActiveJob framework

app/services/make_fresh_downloadable.rb
```diff
@@ -93,13 +93,13 @@ def call(format, filename, html, download_entity, author_string, kwic_text: nil)
 
         formatted_text = kwic_text.gsub("\n", "\r\n") # windows linebreaks
         begin
-          temp_file = Tempfile.new('tmp_kwic_' + download_entity.id.to_s, 'tmp/')
```
damisul, line 96: in a docker container we should use system tmp folder, which is used by default

app/services/make_fresh_downloadable.rb
```diff
@@ -93,13 +93,13 @@ def call(format, filename, html, download_entity, author_string, kwic_text: nil)
...
+          temp_file = Tempfile.new('tmp_kwic_' + download_entity.id.to_s)
           temp_file.puts(formatted_text)
           temp_file.rewind
           temp_file.chmod(0o644) # Set file permissions after writing and rewinding
           dl.stored_file.attach(io: temp_file, filename: filename)
         ensure
-          temp_file.close
+          temp_file&.close
```
damisul, line 102: it crashed here if temp file was not created successfully

spec/controllers/authority_kwic_browser_spec.rb
```diff
@@ -99,31 +99,19 @@
...
-      let(:expression1) { create(:expression, work: work1) }
-      let(:work1) { create(:work) }
-      let(:manifestation1) do
+      let!(:manifestation1) do
         create(
           :manifestation,
-          title: 'First Work',
-          markdown: 'The quick brown fox.',
-          expression: expression1,
-          status: :published
```
damisul, line 110: AI writes extremely verbose specs sometimes. They can be much shorter

abartov (approved):

PR #1500, opened 2026-08-02
Lexicon: arbitrary text→link pairs for works and citations

abartov:
## What

Two related pieces of the lexicon migration verification work:

1. **Editors can link arbitrary words to arbitrary URLs** in a `LexPersonWork` or a `LexCitation`, the way the works' editing views already allowed linking words to *person* entries.
2. **Ingestion preserves inline links it used to discard**, storing them as text/link pairs in that same mechanism.

## Storage

A link pair is `{ text, entry_id }` (a lexicon entry, any type) or `{ text, url }` (anything else).

- `LexCitation` gains a `text_links` JSON column. One set per citation, applied to whichever displayed part contains the text — title, `from_publication`, or `notes`.
- `LexPersonWork` keeps its existing `title_links` / `comment_links` columns and shape; they simply gain `url` targets.

## Rendering

`LexiconHelper#apply_text_links` is now the single implementation behind the works' title/comment rendering, `render_citation`, and the verification cards. URL targets open in a new tab; entry targets stay internal. A citation's title is hyperlinked from `text_links` only when the citation has no `link` of its own, so anchors never nest.

## Editing UI

One shared partial (`lexicon/shared/_text_links_editor`) replaces the two near-identical work partials and serves the new citation section. Its add-form offers an entry autocomplete **or** a URL field, mutually exclusive. `TextLinksConcern` replaces the four duplicated add/remove actions in `PersonWorksController` and backs the new `text_links` / `add_text_link` / `remove_text_link` routes on citations; `javascript:` URLs are rejected.

The person-only restriction on link targets is lifted — it was the direct cause of the dropped links below. A pair whose text no longer occurs in the record is flagged in the list, which matters because the LLM sometimes garbles the surrounding text.

## Detection

**Works** (`ParsePersonWork`): every anchor not pointing at a *person* entry was discarded, losing links to publication and bibliography entries — e.g. on `00156.php`, `01122001.php` (a `text` entry) and `99995017.php` (a `bib` entry) — plus all external URLs. Any resolvable target is now kept; only relative legacy paths that `ProcessLinks` could not resolve are dropped, since storing them would produce links resolved against the entry's own path.

**Citations** (`ParseCitations`): the LLM reports at most one link per citation, so an anchor elsewhere in the record was lost outright or misfiled as the citation's own `link`. Each source `<li>`'s inline anchors are now recorded and joined back to the citation the LLM produced, reusing the link/title matching that already recovered `backup_url`. Anchors already held as the citation's `link`, `backup_url`, or one of its authors are skipped. The prompt also now states that a link on the publication named in `from_publication` is not the work's link, so it stops landing on the title.

Verified against real data — for the `כל העסק מתפרק בכלל` citation of entry 5181 the recovery yields `[{"entry_id"=>6458, "text"=>"שדות ומזוודות"}]`, correctly excluding the author anchor. A 60-file sample also recovers publication links (`כתמי אור`) and external URLs (`shimonbouzaglo.co.il`) that were previously discarded.

No backfill: this applies to future ingests only, by design.

## Drive-by

`notes` is now permitted in `lex_citation_params`. The citation form has rendered a `notes` field all along, but the param was never permitted, so edits silently vanished.

## Test plan

New coverage:
- `spec/helpers/lexicon_helper_spec.rb` — `apply_text_links` with entry/url targets, non-matching and blank-target pairs, escaping; `render_citation` linking inside `from_publication` without nesting anchors in a linked title.
- `spec/services/lexicon/parse_person_work_spec.rb` — non-person entry targets, external URLs in title and comment, unresolvable `.php` and dangling entry ids dropped.
- `spec/services/lexicon/parse_citations_spec.rb` — an unreported inline link becomes a text link; author anchors and the citation's own link are not duplicated.
- `spec/requests/lexicon/citations_spec.rb` — the three new endpoints, plus `notes` persistence.
- `spec/requests/lexicon/person_works_spec.rb` — url pairs, missing target, `javascript:` rejection; the two "entry is not a person → 422" cases now assert the intended 200.
- `spec/system/lexicon/citation_text_links_spec.rb` (`js: true`) — the pair renders on the verification page, appears in the edit modal, and a URL pair can be added through the UI.

Runs: `spec/services/lexicon` (180), `spec/requests/lexicon`, `spec/helpers`, `spec/models/lex_citation_spec.rb`, and the lexicon/citation system specs — all green except 6 pre-existing failures in `spec/requests/lexicon/files_spec.rb`, which are an Elasticsearch flood-stage read-only block from local disk pressure and fail identically on a clean tree. The full suite was not run locally.

### DIFF
--- a/app/controllers/concerns/text_links_concern.rb
+++ b/app/controllers/concerns/text_links_concern.rb
@@ -47,7 +47,10 @@ module TextLinksConcern
     end
 
     url = params[:url].to_s.strip
-    return nil if url.blank? || url.match?(/\Ajavascript:/i) || !url.match?(%r{\A(/|https?://|mailto:)}i)
+    # Same allowlist the ingestion path applies, rather than a denylist of script-bearing
+    # schemes: the url ends up in an href, so anything not known-inert is refused.
+    # (This subsumes the explicit javascript: check of the Copilot autofix it replaces.)
+    return nil unless url.match?(Lexicon::TextLinkExtraction::ALLOWED_URL_PATTERN)
 
     { 'text' => text, 'url' => url }
   end
--- a/app/controllers/concerns/text_links_concern.rb
+++ b/app/controllers/concerns/text_links_concern.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+# Shared add/remove handling for the text→link pairs stored in a JSON column
+# (LexPersonWork#title_links / #comment_links, LexCitation#text_links).
+# A pair is either { 'text' => ..., 'entry_id' => ... } or { 'text' => ..., 'url' => ... };
+# see LexiconHelper#apply_text_links for how they are rendered.
+module TextLinksConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def add_text_link_to(record, field)
+    pair = text_link_pair
+    return head :unprocessable_content if pair.nil?
+
+    links = Array(record.public_send(field))
+    # a given text can only be linked once, since it is substituted by first occurrence
+    record.update!(field => links + [pair]) if links.none? { |link| link['text'] == pair['text'] }
+
+    head :ok
+  end
+
+  def remove_text_link_from(record, field)
+    # Parse the :index param strictly: a non-numeric value ('abc'.to_i == 0) must not
+    # silently delete the first link.
+    index = Integer(params[:index], exception: false)
+    return head :unprocessable_content if index.nil?
+
+    links = Array(record.public_send(field))
+    links.delete_at(index) if index >= 0 && index < links.size
+    record.update!(field => links.presence)
+
+    head :ok
+  end
+
+  # Builds a link pair from the :text plus either the :entry_id or the :url param,
+  # returning nil when the input cannot form a usable pair.
+  def text_link_pair
+    text = params[:text].to_s.strip
+    return nil if text.blank?
+
+    entry_id = params[:entry_id].presence.to_i
+    if entry_id.positive?
+      return nil unless LexEntry.exists?(id: entry_id)
+
+      return { 'text' => text, 'entry_id' => entry_id }
+    end
+
+    url = params[:url].to_s.strip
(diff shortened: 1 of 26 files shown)


abartov:
Addressed in d76c8bff0 (rebased onto the Copilot autofix commit).

**1. `text_links_concern.rb:52` — only `javascript:` was blocked.** Valid, and the more serious of the three. Fixed by switching the controller from a denylist to the same allowlist the ingestion path already used (`http(s)://`, `mailto:`, or a path on this site), hoisted into `Lexicon::TextLinkExtraction::ALLOWED_URL_PATTERN` so both paths share one definition instead of drifting. This subsumes the inline autofix on the same line, which is why the rebase resolved in favour of the constant. Covered by request specs over `javascript:`, `data:`, `vbscript:`, a mixed-case `JaVaScRiPt:`, and `ftp://`, plus a positive case for `/files/lex/...`.

**2. `_text_links_editor.html.haml:20` — N+1 over `entry_id`.** Valid: consolidating the two work partials into one dropped the `index_by` preload the controllers used to do. Restored via `LexiconHelper#text_link_entries`, which builds the map in one query; `apply_text_links` now uses it too, so there is a single preload path.

**3. `_text_links_editor.html.haml:62`/`:63` (and the suppressed one at `:65`) — stacked handlers on AJAX reload.** Not applying this one; the premise doesn't hold for this partial.

The partial is injected with `$('#…-links-list').load(url)`, and jQuery's `load` funnels through `self.html(responseText)` (`jquery3.js:10229`). `html(value)` calls `jQuery.cleanData(getAll(elem, false))` before assigning `innerHTML` (`jquery3.js:6170`), which tears down the event handlers of every element being replaced. The `#{prefix}_url` input and `#{prefix}_entry_autocomplete` field are *inside* the reloaded container, so each reload destroys the old nodes and their handlers, and the script then binds to the freshly-created ones. There is no element that survives a reload holding an old binding, so nothing accumulates — `off().on()` would guard against a scenario that cannot occur here, at the cost of implying one can.

Happy to add it anyway if you'd rather have the belt-and-braces.

PR #1517, opened 2026-08-05
Replace markdown helper button strip with a single action dropdown

abartov:
## Summary

The four markdown-editing views each hand-rolled their own strip of five buttons for the markdown utilities:

| View | Container | Buttons it showed |
|---|---|---|
| `html_file/edit_markdown` | `edit_markdown` | stanza break, add/remove poetry markup, makafim, **tidy footnote dirs** |
| `ingestibles/_form` (full-markdown tab) | `full_markdown` | stanza break, add/remove poetry markup, makafim, **indent** |
| `ingestible_texts/_edit` | `ingestible_edit_texts` | same as above |
| `manifestation/edit` | `manifestation_edit_markdown` | same as above |

The subsets were inconsistent even though `shared/_markdown_utils` already defined handlers for all six actions in every view — drift, not design.

All four now render a single shared partial: a dropdown listing all six actions plus one button that applies the selected one. **The dropdown keeps its selection after execution**, so an action can be repeated by clicking the button again.

## Changes

- **`app/views/shared/_markdown_actions.html.haml`** (new) — the dropdown + apply button. JS hooks are classes (`.js-markdown-action`, `.js-markdown-action-apply`) rather than ids, because the ingestible form renders two markdown editors on one page; the old code used duplicate ids and worked around it by scoping every selector with `container_id`. The `<select>` deliberately carries no `name`, so it never gets submitted with the surrounding forms.
- **`app/views/shared/_markdown_utils.html.haml`** — the six `click` handlers become a `markdown_actions` dispatch table keyed by the dropdown's option values, wired to one delegated click handler scoped to `container_id`. The two byte-identical caret-insertion handlers (`add_stanza_break`, `add_indent`) collapse into `insert_at_caret()`. The action bodies are otherwise unchanged.
- **`ApplicationHelper#markdown_action_options`** — the option list (label → action key).
- The four views drop their button strips and render the partial.
- `application.scss` — `.markdown-actions` flex layout, wrapping since `ingestible_texts/_edit` places the widget in a `col-sm-2`.
- `he.yml` / `en.yml` — `markdown_action` (aria-label) and `apply_markdown_action` (button label).

## Not touched

- `manifestation/edit`'s `#add_image` / `#add_all_images` buttons — they're image-specific and driven by the ddslick `<select>`, not markdown text munging.
- The `#preview` suspicious-bracket highlighting in `_markdown_utils`.

## Test plan

New system spec `spec/system/markdown_action_dropdown_spec.rb` (4 examples, all passing):
- dropdown renders with all six options in order; none of the old button ids remain
- applying "indent" inserts `&emsp;`, the selection survives, and clicking again inserts a second one (the repetition requirement)
- switching the selection dispatches a different action (stanza break, then makafim with its alert)
- the apply button does not submit the surrounding form

Full suite: **2222 examples, 0 failures, 14 pending**.
`rubocop` and `haml-lint` clean on all changed/new files.

Closes by-dey.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1291,3 +1291,21 @@ p.by-genre-name-v02.headline-2-v02 {
   border-radius: 3px;
   background-color: #fff;
 }
+
+// markdown-editing action dropdown + apply button (shared/_markdown_actions)
+.markdown-actions {
+  display: flex;
+  flex-wrap: wrap; // the widget sits in columns as narrow as col-sm-2
+  align-items: center;
+  gap: 6px;
+  margin-bottom: 8px;
+
+  .js-markdown-action {
+    flex: 1 1 140px;
+    min-width: 0;
+  }
+
+  .js-markdown-action-apply {
+    flex: 0 0 auto;
+  }
+}
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,6 +9,19 @@ module ApplicationHelper
     s.force_encoding('UTF-8')
   end
 
+  # Options for the markdown-editing action dropdown (see shared/_markdown_actions).
+  # Each value is the action key dispatched by the JS in shared/_markdown_utils.
+  def markdown_action_options
+    [
+      [t(:add_stanza_break), 'add_stanza_break'],
+      [t(:add_angled_brackets), 'add_angled_brackets'],
+      [t(:remove_angled_brackets), 'remove_angled_brackets'],
+      [t(:minuses_to_makafim), 'minuses_to_makafim'],
+      [t(:indent), 'add_indent'],
+      [t(:tidy_footnote_dirs), 'tidy_footnote_dirs']
+    ]
+  end
+
   def get_intro(markdown)
     lines = markdown[0..2000].lines[1..-2]
     if lines.empty?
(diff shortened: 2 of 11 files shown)


abartov:
Pushed 4b649278d addressing the review comment and adding a centering action.

**Copilot's comment (`replaced` leaking a global):** fixed — `var replaced = 0;` in both `minuses_to_makafim` and `tidy_footnote_dirs`. It was pre-existing code carried over verbatim in the refactor; good catch.

**New action — "מירכוז" (center):**
- with a selection: wraps *precisely* the selected text in a `<center>` tag
- with no selection: wraps the whole line the caret sits on

`insert_at_caret` is now expressed in terms of a more general `replace_range(start, end, txt)`, which the new action reuses to swap out the caret's line; `caret_line_range()` computes the line bounds (guarding the caret-at-offset-0 case, where `lastIndexOf(needle, -1)` would otherwise clamp to 0 and mis-report the line start on a buffer beginning with a newline).

Two new examples in `spec/system/markdown_action_dropdown_spec.rb` cover both branches, asserting the exact resulting buffer so that "precisely the selected text" is distinguishable from "the whole line":

```ruby
select_range(markdown.index('שורה'), 'שורה'.length)
apply_action(I18n.t(:center_text))
expect(find('#markdown').value).to eq("<center>שורה</center> ראשונה\nמילה-מחוברת\n")
```

Full suite: **2224 examples, 0 failures, 14 pending**. `rubocop` and `haml-lint` clean on the changed files, except a file-level `ViewLength` warning on `_markdown_utils.html.haml` (124/100) — it's a JS module living in a HAML template, and the repo already carries much larger ones (`ingestibles/_form` is 448/100). Pronto reports per changed line, so it won't surface there.

PR #1539, opened 2026-08-07
Mobile menu: add all-collections link; keep filter checkboxes beside their labels

abartov:
Two unrelated UI fixes reported together.

## 1. Mobile menu was missing the all-collections list

The mobile sidenav's "יצירות" section had "most read" / "newest" / "list all
works", but not the **"רשימת כל הכותרים"** entry the desktop dropdown has — so
on mobile there was no menu route to `/collections`.

Added it, using the same glyph and label as the desktop dropdown. Also added the
missing `list_all_collections` key to `en.yml` (the desktop menu was already
using it and rendering it untranslated in English).

## 2. Browsing filter checkboxes detached from their labels

Reported symptom: at certain zoom levels the filter pane shows checkboxes with
no label next to them, while the label appears in the next column with no
checkbox.

**Cause** — `label` is `display: inline-block` site-wide (`BY_styles_General.css`),
which makes every label an atomic, unbreakable box. The filter panes lay their
options out in two CSS columns (`.nested-list { column-count: 2 }`), so each
option gets roughly half of an already narrow pane. As soon as a label is wider
than what is left of the line beside its 18px checkbox, the whole label drops to
the next line and the checkbox is stranded alone above it — which reads as a
checkbox belonging to the label above. The column breaker could separate the
pair as well.

**Fix** — each option is now a `.filter-checkbox` row laid out as a flex line:
the label wraps its own text while staying beside its own checkbox, at any zoom
level or label length. The row is also `break-inside: avoid`, so a column break
cannot split it.

Applied to the shared filters partial — used by **/works**, **/authors**,
**/collections** and the **lexicon entries list** — and to the **authority TOC
filter pane** (`authors/_toc_filters`), which had the same markup shape.

Vertical rhythm and spacing are unchanged; only the row's internal layout is.

## Tests

- `spec/requests/mobile_menu_collections_link_spec.rb` — the mobile menu links to
  the collections list, with the same label as the desktop dropdown.
- `spec/system/browse_filter_checkbox_alignment_spec.rb` — for /works, /authors
  and /collections, at four viewport widths, asserts via `getBoundingClientRect`
  that every checkbox shares a line with the first line of its own label and sits
  immediately to its right (RTL). Verified to **fail** on the pre-fix CSS at all
  four widths, including the default 1400px, and pass with the fix.

Also re-ran the neighbouring filter/menu specs: `authors_filter_panel_spec`,
`collections_browse_authorities_filter_spec`, `lexicon/entries_list_filtering_spec`,
`works_browse_authors_filter_spec`, `collections_browse_spec`,
`manifestation_collection_type_filtering_spec`, `author_toc_filters_spec`
(system + request), `author_navbar_mobile_expansion_spec`,
`navbar_works_about_author_i18n_spec` — all green (71 examples).

The full suite (40+ min) was **not** run; it needs your go-ahead.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1561,6 +1561,33 @@ p.by-genre-name-v02.headline-2-v02 {
   color: #907989;
 }
 
+// Filter panes (/works, /authors, /collections, lexicon list) lay their option
+// lists out in CSS columns (.nested-list, column-count: 2), so each option gets
+// only about half the (already narrow) pane. Labels are display: inline-block
+// site-wide, which makes them unbreakable atomic boxes: as soon as a label is
+// wider than what is left of the line beside its 18px checkbox, the whole label
+// drops to the next line and the checkbox is stranded alone on its own line —
+// looking, at a glance, like a checkbox belonging to the label above it. Laying
+// the row out as a flex line instead lets the label wrap its own text while
+// staying beside its checkbox, at any zoom level or label length.
+.filter-checkbox {
+  display: flex;
+  align-items: flex-start;
+
+  // Also keep the row whole when the CSS-column breaker picks a break point.
+  -webkit-column-break-inside: avoid; // Safari, older Chrome
+  page-break-inside: avoid; // Firefox
+  break-inside: avoid;
+
+  > input[type='checkbox'] {
+    flex: none; // never let the checkbox itself be squeezed
+  }
+
+  > label {
+    min-width: 0; // allow the label to shrink below its max-content width
+  }
+}
+
 // In list-view result headers (.headline-2-v02), push the permalink to the
 // left edge of its container (away from the result count) and de-emphasize it
 // by overriding the container's bold with a regular weight.
--- a/app/views/authors/_toc_filters.html.haml
+++ b/app/views/authors/_toc_filters.html.haml
@@ -16,51 +16,58 @@
           %ul
             - fd[:genres].each do |genre|
               %li
-                %input.toc-filter-genre{ type: 'checkbox', value: genre, id: "toc-filter-genre-#{genre}", name: 'toc_filter_genre' }
-                %label{ for: "toc-filter-genre-#{genre}" }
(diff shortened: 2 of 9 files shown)


abartov:
Addressed the review in d1f894906.

**Copilot — `%label` without `for` (`_checkboxes.html.haml`)**
Fixed, and it resolves a bead (`beads_by-brd`) I'd filed for exactly this. The
input already had a stable id, so the label is now bound to it. The one risk was
id uniqueness: `shared/filters/_languages` renders this partial *twice* under the
same `group_name` (`:lang`), relying on the two value sets being disjoint — so
`spec/requests/filter_checkbox_labels_spec.rb` asserts both the label/input
binding and that every generated checkbox id is unique, on all four pages that
render the pane (/works, /authors, /collections, lexicon entries list).
`spec/system/filter_label_click_spec.rb` additionally clicks the *label text* and
checks the filter is applied and then cleared, since that was the behaviour that
didn't work before.

**Copilot — nil dereference in `mobile_menu_collections_link_spec.rb`**
Correct, `link.text` would have raised `NoMethodError` and buried the real cause.
Both the menu lookup and the link lookup now assert presence first, with
explanatory messages.

**haml-lint (CI bot)**
- `_toc_filters.html.haml` lines 20/47/55/61 — these went over 120 chars because
  my `.filter-checkbox` wrapper added two spaces of indentation. Rewrapped the
  four attribute hashes.
- `_tabs.html.haml:178` — added the spaces inside the hash braces.
- `_checkboxes.html.haml` — the new `checkbox_id` local made three consecutive
  silent scripts; merged them into a `:ruby` filter.

The remaining LineLength warnings in `_toc_filters.html.haml` are on pre-existing
lines this PR doesn't touch.

Re-ran: the four specs in this PR plus `author_toc_filters_spec` (system +
request), `authors_filter_panel_spec`, `collections_browse_authorities_filter_spec`,
`lexicon/entries_list_filtering_spec`, `works_browse_authors_filter_spec`,
`collections_browse_spec`, `manifestation_collection_type_filtering_spec` —
73 examples, 0 failures. Full suite still not run.

PR #1545, opened 2026-08-08
Implement pageless authorities (by-7qg)

abartov:
## What & why

Two production `Authority` records — *anonymous* and *various authors* — aren't real creators. They're useful for crediting a text on the read page, but an `Authority#toc` for them would falsely suggest hundreds of works by one prolific author, and there's no meaningful curated content (tags, bio, featured work) to put on such a page.

This adds a `pageless` flag and makes such authorities un-linked and un-listed everywhere.

## Changes

**Model / data**
- Migration adds `pageless` boolean to `authorities`, default `false`, `null: false`.
- `Authority.pageless_ids` / `.pageless_id?` — the id list is tiny and changes very rarely, so it's cached and consulted in memory rather than hitting the DB for every authority mentioned on a page.
- An `after_commit` busts that cache when the flag is toggled, created or destroyed.

  *Note:* the bead suggested preloading the list at server startup. I used a cache-with-bust instead, because a boot-time preload would require a restart after an editor toggles the flag, and wouldn't propagate across Puma workers.
- `scope :not_pageless`; `scope :featurable` now also excludes pageless.

**Helpers (the one chokepoint)**
- `authority_link(authority, text = nil, **html_options)` — a link for normal authorities, the escaped name for pageless ones.
- `link_to_authority(authority, html_options = {}, &block)` — block form for author cards/popups; emits a `span` instead of an `a` so the layout is preserved but nothing is clickable.
- **Every** call site that pointed at `authority_path` now goes through these — ~45 sites across reader views, admin views, `BybeUtils#textify_authorities_and_roles` (which feeds epub/PDF exports and KWIC concordances), and the homepage newsfeed service.

**Listing / discovery**
- `AuthoritiesIndex` `index_scope` excludes pageless → gone from the authorities list *and* from site-wide search results.
- `AuthoritiesAutocompleteIndex` deliberately still includes them: editors must be able to attach works to *anonymous*.
- `featurable` covers homepage carousels, `popular_authors` and `RandomAuthor`.

**Page access**
- `Authority#toc` refuses pageless authorities for non-editors (redirect to `/` with the existing `author_not_available` flash). Editors keep access so they can manage the record.

**i18n** — `pageless` attribute label and `pageless_hint` in both `he` and `en`.

## Setting the flag

The column defaults to `false` and there is **no data migration**. Tick the *Pageless authority* checkbox on the two production records via the authority edit form.

## Decisions taken with @asafbartov during the session

1. TOC page: blocked for non-editors, still visible to editors.
2. Degrade **everywhere**, including admin views (no per-view carve-out).
3. Flag set via admin checkbox only, no name-matching data migration.

A consequence of (2)+(1) worth knowing: with links degraded even on `authors/list` and `authors/edit_toc`, an editor reaches a pageless authority's TOC by URL rather than by clicking. I chose a non-viewer-aware helper deliberately — there are 24-hour fragment caches containing author links (`welcome/_newsfeed`, `_top_authors`, `_authors`, `shared/_author_popup`), and a viewer-aware helper would let an editor's render get cached and served to readers.

## Incidental fix

`welcome/_featured_author_popup` passed `@featured_author.person.id` where an *authority* id was expected (`_featured_author` next to it correctly uses `person.authority`). Converting it to the new helper forced the correction.

## Not touched (flagging, not fixing)

- `authors#print` renders a printable TOC with no publication gate at all today — so it also has no pageless gate. Fixing it properly means also gating unpublished authorities, which is a separate pre-existing issue.
- The V1 API still exposes a `url` for pageless authorities (it would now 302 to home). Removing a field is a public-contract change; say the word if you want it.
- `Authority.cached_count` ("N authors in the project") still counts pageless records. It already counts unpublished ones, so it's an approximation either way.

## Test plan

New tests, each verified to **fail without the corresponding fix**:

- `spec/models/authority_spec.rb` — `not_pageless` / `featurable` scopes, `pageless_ids` / `pageless_id?`, and cache invalidation on toggle-on, toggle-off and destroy (using a real `MemoryStore`, since the test env uses `:null_store`).
- `spec/helpers/application_helper_spec.rb` — 10 examples covering both helpers: link vs. plain text, explicit link text, bare ids, html options passthrough, HTML escaping of the name, and the span-with-classes-but-no-target case.
- `spec/controllers/authors_controller_spec.rb` — `#toc` redirects for anonymous visitors but renders for editors; `#browse` omits a pageless authority from the list (`render_views` + real ES import).
- `spec/controllers/manifestations_controller_spec.rb` — `#read` still *names* a pageless author (the reader needs the credit) but does not link to them.

Specs run (all passing):

| Suite | Result |
|---|---|
| `spec/models/authority_spec.rb`, `spec/helpers/`, `spec/controllers/authors_controller_spec.rb`, `spec/controllers/manifestations_controller_spec.rb` | 404 examples, 0 failures |
| `spec/lib/`, `spec/api/`, `spec/services/`, `spec/requests/` | 1111 examples, 0 failures (9 pre-existing pendings) |
| `spec/controllers/{collections,search,proofs,publications,taggings,admin_controller}_spec.rb` | 215 examples, 0 failures |
| `spec/controllers/admin/`, aboutnesses, involved_authorities, bib, html_file | 96 examples, 0 failures |
| system: whatsnew_page, periodicals_whatsnew, collection_show_authorities, collection_show_orig_lang, periodical_toc_display | 31 examples, 0 failures |

RuboCop and haml-lint are clean on all changed lines (both files have substantial pre-existing offences that I left alone).

**The full suite was NOT run** — it takes 40+ minutes and needs your go-ahead.

Closes by-7qg.


abartov:
Addressed all three review points in 556b5bfc3.

**1. `authors#print` gated like `#read`/`#readmode`**

Followed the `set_manifestation` pattern (`manifestation_controller.rb:1445`): published-or-editor → otherwise flash + redirect to `/`. Extracted the condition into `authority_viewable?`, now shared with `#toc`, and applied to `#print` via a `before_action` declared after `set_author`.

This closes the *unpublished* hole too, not just the pageless one — `#print` previously had no gate whatsoever.

**2. API returns a null `url` for pageless authorities**

`au.pageless? ? nil : authority_url(au)`. The key is still present, just null, so the response shape is unchanged for consumers; documented on the field.

**3. `Authority.cached_count` counts only published**

Now `published.not_pageless.count`.

> One judgement call to confirm: you asked for published-only, and I also excluded pageless. The number is rendered as "N authors in the project" and, in `layouts/application.html.erb`, is the anchor text of a link to `authors_path` — the very list that now excludes pageless authorities. Counting them there would make the label disagree with the list it points at. Say the word if you'd rather it be `published.count` flat.

I also renamed the cache key `au_total_count` → `au_published_count`, so a deploy picks up the corrected figure immediately instead of serving the old inflated number for up to 12 hours.

**Not changed:** `cached_toc_count` (mobile totals) still counts `has_toc` regardless of status — same class of oversight, but you didn't ask for it and it's a different number with different semantics. Happy to fix it in a follow-up bead if you want.

### New tests

- `spec/controllers/authors_controller_spec.rb` — `#print` refuses anonymous visitors and renders for editors, for both pageless *and* unpublished authorities (4 new examples).
- `spec/api/v1/authorities_api_spec.rb` — pageless authority returns a present-but-null `url` with metadata otherwise intact.
- `spec/models/authority_spec.rb` — `cached_count` excludes unpublished, awaiting_first and pageless.

### Specs run (all passing)

| Suite | Result |
|---|---|
| `spec/models/authority_spec.rb`, `spec/controllers/authors_controller_spec.rb`, `spec/api/` | 226 examples, 0 failures |
| `spec/controllers/{manifestations,collections}_controller_spec.rb`, `spec/helpers/`, `spec/requests/`, system: whatsnew_page + collection_show_authorities | 800 examples, 0 failures, 10 pre-existing pendings |

RuboCop clean on all changed lines. The full suite still has not been run — it takes 40+ minutes and needs your go-ahead.

abartov:
Done in 067afaedf.

- `shared/_tabs.html.haml:139` (mobile sidenav header) now uses `Authority.cached_count`, matching the desktop menu a few lines above at `:5`.
- `Authority.cached_toc_count` removed — it had no other caller.

### One orphan left behind, needs your call

`ApplicationController#randomize_authors` (`application_controller.rb:123`) is the only remaining reference to a `cached_toc_count` name, and it is thoroughly dead:

- **no callers** anywhere in `app/`, `lib/` or `spec/`;
- it calls `Person.cached_toc_count` and `Person.has_toc` — **neither has ever existed on `Person`** (both are/were `Authority` methods), so it would raise `NoMethodError` if anything did call it;
- it carries a `# TODO: fix bug where only one author is retrieved by above block` comment.

Removing `Authority.cached_toc_count` doesn't change its status — it was already broken in exactly the same way, since it referenced `Person`, not `Authority`. I've left it alone rather than delete code you didn't ask me to touch.

**Should I remove `randomize_authors` in this PR, or would you rather it be a separate bead?**

### Test

New `spec/requests/mobile_menu_totals_spec.rb` — asserts the mobile totals string matches `Authority.cached_count`, and that unpublished and pageless authorities aren't counted. Both examples verified to fail against the old `has_toc`-based count.

### Specs run

`spec/requests/`, `spec/models/authority_spec.rb`, `spec/controllers/authors_controller_spec.rb`, `spec/system/header_stats_mobile_stat_spec.rb`, `spec/system/whatsnew_page_spec.rb` — **661 examples, 0 failures**, 10 pre-existing pendings.

RuboCop and haml-lint clean on changed lines. Full suite still not run.

abartov:
Removed in 1df51968b.

`ApplicationController#randomize_authors` is gone. Nothing else referenced it, and nothing referenced `cached_toc_count` afterwards either (the only remaining mention is a comment in the new mobile-totals spec explaining why the count changed).

RuboCop on `application_controller.rb`: 26 → 21 offences, i.e. the deletion only removed pre-existing ones and introduced none.

### Specs run

Whole `spec/controllers/` and `spec/requests/` trees — **1408 examples, 0 failures**, 9 pre-existing pendings.

The full suite still has not been run; happy to kick it off whenever you want the ~40 minutes spent.

PR #1546, opened 2026-08-08
Mobile browse lists: show the list first, collapse filters behind a toggle

abartov:
## Problem

On mobile, `/works`, `/authors` and `/collections` render the sort/filter panel as the first column. Once the Bootstrap columns stack (below the `md` breakpoint, 768px) that panel occupies the entire first screenful and the list itself starts off-screen.

## Changes

- **`app/views/shared/_mobile_filter_toggle.html.haml`** (new): the single toggle button (`#mobile_filter_btn`), rendered from each browse page's header partial so it sits above the list.
- **`app/assets/stylesheets/application.scss`**: below 768px, `#sort_filter_panel.mobile-collapsible-filters` is collapsed and `.mobile-filters-open` reveals it. The button is hidden at ≥768px — `.author-page-top-sort-mobile` reveals it at 991px, but between 768 and 991 the columns are still side by side and the panel is permanently visible, so the toggle would be moot there.
- **`app/assets/javascripts/application.js`**: delegated `#mobile_filter_btn` handler (toggles the class, `aria-expanded`, and the button label) plus a `closeMobileFilters()` helper. Delegation matters because `browse.js.erb` replaces the panel's contents on every filter submit.
- **`_browse_filters.html.haml`** (works/authors/collections): the existing "apply selected filter" button now calls `closeMobileFilters()` instead of applying inline `display` styles to the panel and list.
- **`collections/browse.html.haml`**: collections already had a `#mobile_filter_btn` whose handler swapped the list out for the panel; that bespoke handler is dropped in favour of the shared one.
- **locales**: new `hide_sort_and_filter` key (he + en) for the toggled-open label.

`/anthologies` has its own `#sort_filter_toggle` mechanism and is deliberately left alone; the new CSS is scoped by the `.mobile-collapsible-filters` class so it can't reach that page.

## Test plan

New `spec/system/browse_mobile_filter_toggle_spec.rb` (js), covering all three pages at 375x812:
- the list card is inside the viewport, the filter panel is hidden, the toggle is visible;
- clicking the toggle reveals the panel and flips the label/`aria-expanded`, clicking again re-hides it;
- at 1400x900 the panel stays open and the toggle is hidden.

Verified the specs fail (7/7) with the app changes stashed and pass with them applied.

Specs run: `browse_mobile_filter_toggle_spec.rb` (7 examples, 0 failures) plus the neighbouring browse specs — `works_mobile_sorting_spec.rb`, `authors_filter_panel_spec.rb`, `collections_browse_authorities_filter_spec.rb`, `browse_filter_checkbox_alignment_spec.rb`, `browse_permalink_spec.rb`, `filter_label_click_spec.rb`, `anthology_browse_spec.rb` (35 examples, 0 failures), and `collections_browse_spec.rb`, `works_browse_authors_filter_spec.rb`, `authors_controller_spec.rb`, `collections_controller_spec.rb` (156 examples, 0 failures).

**The full suite was not run** (40+ min) and needs your approval.

Closes beads_by-wdc

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -31,17 +31,32 @@ function isMobile() {
 // Browse lists (/works, /authors, /collections): on narrow screens the
 // sort/filter panel is collapsed by CSS so the list starts in the first
 // screenful, and #mobile_filter_btn is the single control that reveals it.
-function closeMobileFilters() {
-  $('#sort_filter_panel').removeClass('mobile-filters-open');
+// The open state lives on <body> because the panel and the buttons that drive
+// it sit in different subtrees (the buttons are in the fixed page header).
+function setMobileFiltersOpen(isOpen) {
+  $('body').toggleClass('mobile-filters-open', isOpen);
   var btn = $('#mobile_filter_btn');
-  btn.attr('aria-expanded', 'false').text(btn.data('show-label'));
+  btn.attr('aria-expanded', isOpen ? 'true' : 'false')
+     .text(isOpen ? btn.data('hide-label') : btn.data('show-label'));
 }
 
-// Delegated, so the handler survives the AJAX re-render of the filter panel.
+function closeMobileFilters() {
+  setMobileFiltersOpen(false);
+}
+
+// Delegated, so the handlers survive the AJAX re-render of the filter panel.
 $(document).on('click', '#mobile_filter_btn', function() {
-  var isOpen = $('#sort_filter_panel').toggleClass('mobile-filters-open').hasClass('mobile-filters-open');
-  $(this).attr('aria-expanded', isOpen ? 'true' : 'false')
-         .text(isOpen ? $(this).data('hide-label') : $(this).data('show-label'));
+  setMobileFiltersOpen(!$('body').hasClass('mobile-filters-open'));
+});
+
+// #apply_mobile_filters submits the filter form (via its `form` attribute) and
+// collapses the panel again, putting the freshly filtered list back on screen.
+$(document).on('click', '#apply_mobile_filters', function() {
+  if (typeof resetPagination === 'function') {
+    resetPagination();
+  }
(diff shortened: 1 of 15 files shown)


abartov:
Addressed the Copilot review and added the follow-up requested in review:

**Copilot's point (both comments were the same issue)** — the wrapper, not just the button, is now hidden at ≥768px. The partial's wrapper carries a dedicated `.mobile-filter-toggle` class alongside `.author-page-top-sort-mobile`, and `application.scss` hides `.author-page-top-sort-mobile.mobile-filter-toggle` there, so no empty flex item with margins is left in the authors/works headers between 768–991px. Other `.author-page-top-sort-mobile` users are untouched. Covered by the desktop example, which now asserts the wrapper itself is hidden.

**New: the mobile 'apply filters' button now floats beside the toggle.** It used to sit at the foot of the filter panel — several screenfuls below on a phone. It now renders in the shared toggle partial next to `#mobile_filter_btn`; that row is inside the fixed `#header`, so both buttons stay on screen for the whole scroll through the filters (verified in a spec that scrolls 600px and checks both bounding rects are inside the viewport). An HTML `form=` attribute keeps it submitting the filters form it is no longer nested in, and its click handler moved to a delegated one in `application.js` — the button is no longer re-rendered with the panel, so a per-render binding would have accumulated one handler per AJAX filter submit.

Open state moved from a class on `#sort_filter_panel` to one on `<body>`, since the panel and the buttons driving it are now in different subtrees.

One behaviour change worth flagging: the apply button was `.mobile-only` (≤991px) and is now ≤767px only. Between 768–991px the columns are side by side, the panel is permanently visible and filters auto-apply on change, so the button had no job there.

Specs: `browse_mobile_filter_toggle_spec.rb` now 13 examples, 0 failures. Re-ran the neighbouring browse specs — 191 examples, 0 failures. Full suite still not run.

